PR #24, opened 2022-12-13
Docs for 0.4

lovromazgon:
This PR refactors the structure of our docs and introduces the root element `Features` which groups several existing docs taken from the Conduit repository (pipeline semantics, metrics, API, UI) and adds a document about dead-letter queues.


docs/configuration/pipeline-configuration-files.mdx
```diff
@@ -1,6 +1,5 @@
 ---
 title: 'Pipeline Configuration Files'
-slug: 'pipeline-configuration-files'
```
hariso, line 3: Default I guess?

docs/connectors/behavior.mdx
```diff
@@ -1,7 +1,5 @@
 ---
 title: "Connector Behavior"
-sidebar_label: "Behavior"
-slug: "behavior"
```
hariso, line 4: Will the slug now be `connector-behavior`? If so, do we maybe have any backlinks to this?

docs/features/dead-letter-queue.mdx
```diff
@@ -0,0 +1,87 @@
...
+
+Below is a visual representation a sliding window monitoring 5 records with the
+nack threshold set to 2. Notice that once the threshold is exceeded the pipeline
+stops.
+
+![nack window animation](/images/dlq-window-animation.gif)
+
+## Configuring a dead-letter queue
+
+By default a Conduit pipeline doesn't have a dead-letter queue. You can enable
```
hariso, line 48: > By default a Conduit pipeline doesn't have a dead-letter queue.

Nitpick: explicitly stating that this means that the first error stops a pipeline would be nice.

hariso (approved):
Great work!

docs/configuration/pipeline-configuration-files.mdx
lovromazgon: Yes, the default is the filename. I think it's simpler if we just use that, no need to open the file to know how it needs to be addressed.

docs/connectors/behavior.mdx
lovromazgon: No, this one will stay as it is (see filename). It will change for a couple of other documents in this directory, but I think I changed all links that point toward those.

docs/features/dead-letter-queue.mdx
lovromazgon: Good idea!

hariso: Reads great!

docs/configuration/pipeline-configuration-files.mdx
hariso: What I like about explicitly setting slugs is that no "docs refactoring" will affect URLs. But what you said makes sense too.

lovromazgon: Well there's always pro's and con's 😄

docs/features/dead-letter-queue.mdx
```diff
@@ -0,0 +1,91 @@
...
+Conduit can monitor a fixed number of last records and track which ones are
+acked or nacked. As new records are processed, they are added into the window
+while the oldest ones are discarded to make room for them. This allows the
+window to "slide" through records, always containing the latest records. Conduit
+keeps track of the number of nacks in the sliding window and stops the pipeline
+if the threshold gets exceeded. This can be useful for ensuring that the
+pipeline is operating correctly and effectively.
+
+Below is a visual representation a sliding window monitoring 5 records with the
+nack threshold set to 2. Notice that once the threshold is exceeded the pipeline
```
maha-hajja, line 41:
```suggestion
nack threshold set to 2. Notice that once the threshold is exceeded, the pipeline
```

maha-hajja (commented):
some mini nitpicks, and loverly work on the new documents!

maha-hajja (approved):
lgtm

PR #26, opened 2023-02-01
update last version to 0.5.0

maha-hajja:


lovromazgon (approved):
LGTM, but maybe we should wait before we merge to see if marketing wants us to prepare a tweet and/or blog post before.

PR #27, opened 2023-02-03
fix typo in pipeline configs example

ahmeroxa:
The CLI command for using config files is missing an 's'.


jayjayjpg (approved):
This reads great, thank you for the super quick fix! :sparkles:

PR #28, opened 2023-04-05
Document pipeline configuration files v2.0

lovromazgon:
This PR changes the pipeline configuration file documentation to explain version 2.0 and adds an explanation for each field. The existing documentation is broken up into 3 separate chapters and slightly expanded.


docs/pipeline-configuration-files/provisioning.mdx
```diff
@@ -0,0 +1,61 @@
...
+**_Note_**: Connector IDs and processor IDs will be prefixed with the parent ID. In case of a connector this means the
+pipeline ID will be attached as a prefix to the connector ID (e.g. `pipelineID:connectorID`). Similarly, a pipeline
+processor's ID will be prefixed with the pipeline ID (e.g. `pipelineID:processorID`) and a connector processor's ID will
+be prefixed with the connector ID (e.g. `pipelineID:connectorID:processorID`). This ensures that connector and processor
+IDs are globally unique as long as the pipeline ID is unique.
+
+## Pipelines Immutability
+
+Pipelines provisioned by configuration files are **immutable**, any updates needed on a provisioned pipeline have to be
+done through the configuration file. You can still stop or start the pipeline through the UI or API.
```
hariso, line 40: Nitpick: it feels that "immutable" is not quite technically correct, given that they are mutable, but the changes are restricted to files.

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -0,0 +1,272 @@
...
+- **Data Type**: Mapping node
+- **Required**: No
+- **Default**: See sub-fields
+- **Description**: This node contains the dead-letter queue configuration.
+
+#### plugin
+
+- **Data Type**: String
+- **Required**: No
+- **Default**: `builtin:file`
```
hariso, line 133: Wasn't it the log connector?

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -0,0 +1,272 @@
...
+- **Allowed Values**: `stopped`, `running`
+- **Description**: This field controls the status of the pipeline at startup. If set to `running` the pipeline will be
+  automatically started, if set to `stopped` the pipeline will be provisioned without being started.
+
+### name
+
+- **Data Type**: String
+- **Required**: No
+- **Default**: Same as pipeline [`id`](#id)
+- **Description**: Human readable name for the pipeline. Needs to be unique across all pipelines.
```
hariso, line 113: Can you please remind me why it has to be unique (when we already have the ID), especially because connector names for example do not need to be unique? I vaguely recall something about our platform and fetching pipelines.

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -0,0 +1,272 @@
...
+- **Default**: None
+- **Description**: ID of the [`processor`](#processor). It should uniquely identify the processor inside the pipeline.
+  If multiple processors inside the pipeline use the same ID, provisioning will fail.
+
+### type
+
+- **Data Type**: String
+- **Required**: Yes
+- **Default**: None
+- **Description**: Defines the processor type.
```
hariso, line 256: Would be good to also say the available types or where to find the available types.

docs/pipeline-configuration-files/provisioning.mdx
maha-hajja: we could say:
```suggestion
Pipelines provisioned by configuration files are **immutable** through the API, any updates needed on a provisioned pipeline have to be
done through the configuration files. You can only stop or start the pipeline through the UI or API.
```

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -0,0 +1,272 @@
...
+          field: ${ENV_VAR}     # You can use environment variables by wrapping them in a dollar sign and curly braces ${}
+    dead-letter-queue:          # Dead-letter queue (DLQ) configuration
+      plugin: "builtin:file"    # DLQ Connector plugin
+      settings:                 # A map of configuration keys and values for the plugin (specific to the chosen plugin)
+        path: "./dlq.out"
+      window-size: 5            # DLQ nack window size
+      window-nack-threshold: 2  # DLQ nack window threshold
+```
+
+## version
```
maha-hajja, line 65: love these! this whole file is super detailed and on point.. nice work!

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -0,0 +1,272 @@
...
+- **Default**: None
+- **Description**: This node contains the configuration of a single pipeline.
+
+### id
+
+- **Data Type**: String
+- **Required**: Yes
+- **Default**: None
+- **Description**: ID of the [`pipeline`](#pipeline). It should uniquely identify the pipeline across all pipelines. If
+  multiple pipelines use the same ID, none of them will be provisioned.
```
maha-hajja, line 95: none or the first one?

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -0,0 +1,272 @@
...
+
+**Warning**: Changing this property will cause all connectors to start from the beginning.
```
maha-hajja, line 97:
```suggestion
**Warning**: Changing this property will cause all this pipeline's connectors to start from the beginning when Conduit is restarted.
```

docs/pipeline-configuration-files/specifications.mdx
lovromazgon: Yes, good catch!

docs/pipeline-configuration-files/specifications.mdx
lovromazgon: IIRC it's because we are using the pipeline name in metrics to distinguish them, since names are provided by the user and IDs were not at the time we designed this (they still are not if the pipeline is created through the API). So to prevent mixing metrics between multiple pipelines they need to have unique names.

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -0,0 +1,272 @@
...
+- **Description**: Human readable name for the pipeline. Needs to be unique across all pipelines.
+
+### description
+
+- **Data Type**: String
+- **Required**: No
+- **Default**: None
+- **Description**: Human readable description of the pipeline.
+
+### dead-letter-queue
```
maha-hajja, line 122: I think it's better to have DLQ below the connectors and processors, since they're more important and usually come first

docs/pipeline-configuration-files/provisioning.mdx
lovromazgon: Right, I'll try to rephrase it 👍

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -0,0 +1,272 @@
...
+
+### id
+
+- **Data Type**: String
+- **Required**: Yes
+- **Default**: None
+- **Description**: ID of the [`connector`](#connector). It should uniquely identify the connector inside the pipeline.
+  If multiple connectors inside the pipeline use the same ID, provisioning will fail.
+
+**Warning**: Changing this property will cause the connector to start from the beginning.
```
maha-hajja, line 190:
```suggestion
**Warning**: Changing this property will cause the connector to start from the beginning when Conduit is restarted.
```

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -0,0 +1,272 @@
...
+
+### type
+
+- **Data Type**: String
+- **Required**: Yes
+- **Default**: None
+- **Allowed Values**: `source`, `destination`
+- **Description**: Defines if the connector is a source or a destination. A working pipeline needs at least one source
```
maha-hajja, line 198:
```suggestion
- **Description**: Defines if the connector is a source or a destination. A running pipeline needs at least one source
```

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -0,0 +1,272 @@
...
+- **Required**: No
+- **Default**: Same as connector [`id`](#id-1)
+- **Description**: Human readable name for the connector.
+
+### settings
+
+- **Data Type**: Mapping node
+- **Required**: Yes
+- **Default**: None
+- **Description**: A map of configuration keys and values for the connector plugin. The values in this map depend on the
```
maha-hajja, line 224: we could let the user know where to get the needed settings (checking the config table from the connector docs)

docs/pipeline-configuration-files/specifications.mdx
maha-hajja: oh didn't notice this change! yeah makes sense

docs/pipeline-configuration-files/getting-started.mdx
```diff
@@ -0,0 +1,73 @@
...
+        plugin: builtin:file
+        settings:
+          # write data to example.out
+          path: ./example.out
+          # output the raw payload as a string
+          sdk.record.format: template
+          sdk.record.format.options: '{{ printf "%s" .Payload.After }}'
+```
+
+Now start Conduit and notice the log line saying that the pipeline `file-to-file` was created:
```
simonl2002, line 57: I'd tweak this slightly. Maybe say something like `Now start Conduit. You should see a log line saying...`

docs/pipeline-configuration-files/provisioning.mdx
```diff
@@ -0,0 +1,61 @@
...
+title: 'Provisioning'
+sidebar_position: 1
+---
+
+## Pipelines Directory
+
+When Conduit starts, it will search for all files ending with `.yml` or `.yaml` in folder `./pipelines` and its
+sub-folders. It will parse the configuration files and provision the pipelines defined in them.
+
+If you want Conduit to search a different directory, or want to provision a single file, you can specify the CLI flag
```
simonl2002, line 11: This is implying that you can point `-pipelines.path` to a single file? Is that the case?

docs/pipeline-configuration-files/provisioning.mdx
```diff
@@ -0,0 +1,61 @@
...
+If the directory does not exist, Conduit will fail with the error `"pipelines.path" config value is invalid`.
+
+## Ignored errors
+
+Conduit will try to provision all pipelines regardless of the errors it may encounter while going through the
+configuration files. Invalid configuration files will be ignored (e.g. files that miss a mandatory field or have an
+invalid configuration value) and won't impact the provisioning of valid configuration files. Conduit will log a warning
+for every error it encounters during this process.
+
+You need to be careful when choosing IDs for your pipelines, connectors and processors. Each pipeline needs to get a
```
simonl2002, line 27: Feels like this deserves it's own heading rather than being buried under "ignored errors." It is a pretty important detail.

docs/pipeline-configuration-files/provisioning.mdx
lovromazgon: Yup, that's right!

simonl2002: yeah I tested it myself and came back to just delete this.

docs/pipeline-configuration-files/specifications.mdx
lovromazgon: I added a link to the processor documentation that we have right now, but we don't have a full list of processors, we just point to the [godoc](https://pkg.go.dev/github.com/conduitio/conduit/pkg/processor/procbuiltin) there. I think we can and should do better by having a full list of processors and their capabilities, similar to the [kafka connect docs](https://docs.confluent.io/platform/current/connect/transforms/overview.html). That's actually what [this issue](https://github.com/ConduitIO/conduit/issues/966) was meant for.

lovromazgon:
Thanks for your reviews! I _think_ I addressed the comments so far, let me know if you see more opportunities to improve the doc 🙂 @maha-hajja @hariso @simonl2002

docs/pipeline-configuration-files/provisioning.mdx
```diff
@@ -3,59 +3,106 @@ title: 'Provisioning'
...
+`pipelineID:processorID`) and a connector processor's ID will be prefixed with
+the connector ID (e.g. `pipelineID:connectorID:processorID`). This ensures that
+connector and processor IDs are globally unique as long as the pipeline ID is
+unique.
+
+## Environment variables
+
+Conduit injects environment variables in pipeline configuration files. To
+specify a value using an environment variable, wrap the name of the variable in
+`${}` (e.g. `${MY_ENV_VAR}`). Any value in the configuration can be injected
```
maha-hajja, line 64:
```suggestion
`${}` (e.g. `${MY_ENV_VAR}`). Any value in the configuration file can be injected
```

maha-hajja (approved):
lgtm

hariso (approved):
Great work Lovro, especially the specs!

PR #29, opened 2023-04-06
Connector Output Formats

hariso:


docs/connectors/output-formats.mdx
```diff
@@ -0,0 +1,183 @@
...
+
+:::note
+
+Configuring different output formats is possible for connectors which use Conduit SDK's default
+middleware. Please consult the connector's documentation to confirm the feature is available.
+
+:::
+
+The output format is configured at the level of a destination connector. This implies that different connectors in the
+same or different pipelines can use different formats, which provides additional flexibility.
```
lovromazgon, line 20: I'd keep it simple.
```suggestion
same pipeline can use different formats.
```

docs/connectors/output-formats.mdx
```diff
@@ -0,0 +1,183 @@
...
+
+None.
+
+### `debezium/json`
+
+Generates output compatible with Debezium connectors.
+
+#### **Options**
+
+##### `debezium.schema.name`
```
lovromazgon, line 49: We could mention that it's optional to define this.

docs/connectors/output-formats.mdx
```diff
@@ -0,0 +1,183 @@
...
+          path: /tmp/file-destination.txt
+```
+
+If you run Conduit and then `tail` the destination file, you'll see something like this:
+
+```shell
+employee ID 636561854
+employee ID 1848322815
+employee ID 123675258
+...
```
lovromazgon, line 182: I'd lose the 3 dots.

lovromazgon (approved):
Nice 👍 Just some nitpicks.

PR #30, opened 2023-04-08
Docs for built-in processors

hariso:
Fixes https://github.com/ConduitIO/conduit/issues/966.

I've used the comments in the `procbuiltin` a great deal. The docs in Conduit mention raw keys/payloads with schema. While a concept of schema exists in Conduit, it's nowhere used. Because of that, I've removed the mention of "schema" here.


docs/processors/builtin.mdx
```diff
@@ -0,0 +1,112 @@
...
+  Sets the behavior to "include" or "exclude" the record based on the result of the condition. Possible values
+  are: `include`, `exclude`.
+- `condition` (Required, XPath expression)
+  An XPath query expression that the user defines to forward or drop a record on its results.
+- `missingornull` (Optional)
+  Defines how to handle the record in the event the fields the query would use don't exist. Possible values
+  are: `fail`, `include`, `exclude`. Defaults to `fail`.
+- `exists` (Optional)
+  Gives the user a chance to define an existence query for a given filter. Possible values
+  are: `fail`, `include`, `exclude`. Defaults to `fail`.
```
hariso, line 47: I've copied this from the docs in Conduit, it's a bit unclear, at least to me. Once I'm done with the rest, I'll try add more details and add an example.

docs/processors/builtin.mdx
```diff
@@ -0,0 +1,112 @@
...
+
+Below is a list of available built-in processors in Conduit. An example usage of a built-in processor is available in
+[`extract-field-transform.sh`](https:github.com/ConduitIO/conduit/blob/main/examples/processors/extract-field-transform.sh).
+The script will run Conduit with a pipeline containing the built-in `extractfieldpayload` processor.
+
+### extractfieldkey
+
+- If the key is raw and has a schema attached, extracts the field and uses it to replace the entire key.
+- If the key is raw and has no schema, returns an error (not supported).
+- If the key is structured, extracts the field and uses it to replace the entire key.
```
hariso, line 20: I copied this from the docs in Conduit, but it doesn't appear it's true: https://github.com/ConduitIO/conduit/blob/71c6737846700e24a0cdb04c4fe1a890794b9276/pkg/processor/procbuiltin/extractfield.go#L72-L76.

docs/processors/builtin.mdx
```diff
@@ -0,0 +1,112 @@
...
+- `backoffRetry.count`
+  (Optional, integer) Maximum number of retries when backing off following an error. Defaults to 0.
+- `backoffRetry.factor` (Optional, integer)
+  The multiplying factor for each increment step. Defaults to 2.
+- `backoffRetry.min` (Optional, [duration string](https://pkg.go.dev/time#ParseDuration))
+  Minimum waiting time before retrying. Defaults to `100ms`.
+- `backoffRetry.max` (Optional, [duration string](https://pkg.go.dev/time#ParseDuration))
+  Maximum waiting time before retrying. Defaults to `5s`.
+
+### insertfieldkey
```
hariso, line 93: TODO: describe the behavior when the field already exists.

docs/processors/builtin.mdx
```diff
@@ -0,0 +1,234 @@
...
+---
+
+:::note
+Note that all built-in processors that operate on the payload actually operate on `Record.Payload.After`.
+If you need to manipulate the field `Record.Payload.Before` you can use
+a [JavaScript processor](./javascript-processors).
+:::
+
+Below is a list of available built-in processors in Conduit. An example usage of a built-in processor is available in
+[`extract-field-transform.sh`](https:github.com/ConduitIO/conduit/blob/main/examples/processors/extract-field-transform.sh).
```
lovromazgon, line 13: The link seems to be broken. Although, I'd rather point to the [configuration file](https://github.com/ConduitIO/conduit/blob/main/examples/processors/pipeline-extract-field-transform.yml) instead of the shell script. Or we just link to the [folder](https://github.com/ConduitIO/conduit/tree/main/examples/processors) and ideally add a readme in there that describes the example (essentially we would just copy over the comments from the shell script).

docs/processors/getting-started.mdx
```diff
@@ -59,109 +60,3 @@ The request to create a processor is described in [api.swagger.json](https://git
 
 Conduit provides a number of built-in processors, such as filtering fields, replacing them, posting payloads to HTTP endpoints etc.
 Conduit also provides the ability to write custom processors in JavaScript.
```
lovromazgon, line 62: We could link here to the built-in processors and JavaScript processor page.

docs/processors/getting-started.mdx
```diff
@@ -1,6 +1,7 @@
 ---
 title: 'Getting Started with Processors'
```
lovromazgon, line 2: It seems that this was written when we didn't yet have configuration files since we just mention how processors are created in the API. I think it would be nice to rename "Adding and configuring a processor" to something like "Creating processors through the API" and add a similar section for the configuration file. 

Sidenote: this can be done in a separate PR, as it doesn't seem in scope of this one, although I won't stop you if you want to squeeze it in.

lovromazgon (commented):
I think it's a great start, in the future it would be great to also have examples for each processor. I think that is a bigger piece of work though and should be done separately.

docs/processors/builtin.mdx
hariso: I like the idea of linking to the config. file directly a bit more, since we may add more examples to the folder.

docs/processors/builtin.mdx
```diff
@@ -0,0 +1,234 @@
...
+---
+
+:::note
+Note that all built-in processors that operate on the payload actually operate on `Record.Payload.After`.
+If you need to manipulate the field `Record.Payload.Before` you can use
+a [JavaScript processor](./javascript-processors).
+:::
+
+Below is a list of available built-in processors in Conduit. An example usage of a built-in processor is available in
+[`extract-field-transform.sh`](https://github.com/ConduitIO/conduit/blob/main/examples/processors/pipeline-extract-field-transform.yml).
```
lovromazgon, line 13:
```suggestion
[`pipeline-extract-field-transform.yml`](https://github.com/ConduitIO/conduit/blob/main/examples/processors/pipeline-extract-field-transform.yml).
```

docs/processors/builtin.mdx
```diff
@@ -0,0 +1,234 @@
...
+
+  Replacement value. Defaults to the field type's zero value (e.g. empty string for strings, 0 for integers etc.)
+
+### maskfieldpayload
+
+Works in the same way as [maskfieldpayload](#maskfieldpayload), except that it's applied to `Record.Payload.After`.
+
+### parsejsonkey
+
+Transforms the record's key into structured data, i.e. a JSON object.
```
maha-hajja, line 154: I think this processor and `parsejsonpayload` need more details, maybe something like:
```suggestion
Transforms the record's raw data key (has to be JSON formatted) into structured data, i.e. a JSON object.
```

docs/processors/builtin.mdx
hariso: :facepalm:

docs/processors/getting-started.mdx
```diff
@@ -27,7 +28,20 @@ needs to happen for different sources or destinations, you have two options:
...
 
-## Adding and configuring a processor
+## Creating a processor through a pipeline configuration file
+
+Creating a processor through a pipeline configuration file can be done as below:
+```yaml
+- id: extract-name
+  type: extractfieldpayload
+  settings:
+    field: name
```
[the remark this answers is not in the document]
hariso: Linking to the pipeline config docs sounds good to me. I put just the processor part because that's what I'd at least be interested in knowing, straight to the point. It's also analogous to the API example below. But I don't mind adding a whole pipeline either. Now that I think of it, I can probably just link to https://github.com/ConduitIO/conduit/blob/main/examples/processors/pipeline-extract-field-transform.yml.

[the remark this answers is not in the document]
lovromazgon: My goal is to let the user know where that processor part fits in. Similarly to how the [dead letter queue](https://conduit.io/docs/features/dead-letter-queue#configuring-a-dead-letter-queue) docs skip the irrelevant parts but still show you where the DLQ config is defined.

Maybe my example could be split into 2 examples, one for a connector processor and another for a pipeline processor, this would allow us to keep both very simple. For example:

```
version: 2.0
pipelines:
  - id: example-pipeline
    connectors:
      # define source and destination connectors
      # ...
    processors:
      - id: extract-name
        type: extractfieldpayload
        settings:
          field: name
```

docs/processors/builtin.mdx
maha-hajja: @hariso looks like you didn't notice the comment, we need to update the payload processor too 👀

hariso: @maha-hajja Thinking again about this... IIUC, the `parsejson` pair of processors will parse the JSON even when the data is structured, which means that we don't need to say "transform the record's **raw data** key/payload.after...". Is my understanding correct? I agree that the note about that it has to be JSON formatted should be there.

maha-hajja: @hariso if the data is already structured then it's just returned as is

PR #34, opened 2023-04-14
Processor Concurrency

lovromazgon:
This PR adds documentation for Parallel processors.

Closes https://github.com/ConduitIO/conduit/issues/956.


docs/processors/parallel-processors.mdx
```diff
@@ -0,0 +1,55 @@
...
+order, the resulting order of records coming out of the processor is guaranteed
+to be unchanged.
+
+:::
+
+For instance, consider a pipeline that needs to process 100 records per second.
+One of the processors in the pipeline needs to retrieve data from a server, and
+this operation typically takes 1 second to complete. Since this operation is
+executed for every record, sequential processing won't keep up with the stream.
+However, by configuring the processor to use 200 parallel workers, we can
```
maha-hajja, line 24: why 200? the parallel 100 workers should finish processing in 1 second too, unless you intentionally wanted a higher number, then I think 200 is a bit confusing for the example, and the user might think that they need double the number of records, wdyt?

docs/processors/parallel-processors.mdx
```diff
@@ -0,0 +1,55 @@
...
+larger bursts.
+
+## Configuration
+
+You can configure a processor to run concurrently by specifying the number of
+parallel workers. By default, the number of workers is set to 1, which means
+that records are processed sequentially.
+
+Here is an example pipeline configuration that configures the `js` processor to
+run in parallel using 10 workers:
```
maha-hajja, line 36: is there some kind of max number of workers?

docs/processors/parallel-processors.mdx
```diff
@@ -0,0 +1,55 @@
+---
+title: 'Parallel Processors'
+sidebar_position: 3
+---
+
+Conduit allows you to configure a processor to process records in parallel. This
```
simonl2002, line 6: Maybe we actually mention the parameter.
So something like.
"... to process records in parallel using the `workers` parameter." 

The second sentence feels a little redundant. It is basically just defining parallelism.

docs/processors/parallel-processors.mdx
```diff
@@ -0,0 +1,55 @@
+---
+title: 'Parallel Processors'
```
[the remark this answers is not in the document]
simonl2002: Further down we're actually using "concurrency." Maybe we should pick one or the other. I actually like "Processor Concurrency"

docs/processors/parallel-processors.mdx
```diff
@@ -0,0 +1,55 @@
...
+:::tip Ordering guarantee
+
+When processors are configured to process records concurrently, they guarantee
+that the order of records won't change. Although records may be processed out of
+order, the resulting order of records coming out of the processor is guaranteed
+to be unchanged.
+
+:::
+
+For instance, consider a pipeline that needs to process 100 records per second.
```
simonl2002, line 20: I think we can just remove "For instance,"

docs/processors/parallel-processors.mdx
simonl2002, line 24: I'd replace "parallel workers" => "workers" everywhere.

lovromazgon: I can change the example to 100 to make it clearer. I wanted to show that you want to configure a bit more workers to have some room for higher loads. But I can try mentioning this at the end of the example.

docs/processors/parallel-processors.mdx
lovromazgon: Yes, physics will prevent you from spawning an infinite number of workers 😄 Jokes aside, the only limit right now is the max value we can store in an `int`, on 64-bit systems that's `9223372036854775807`. I'm sure bad things would happen at far lower numbers though, so effectively there's no real limit. We should probably have one!

maha-hajja: yeah makes sense! was just thinking about the users and that they'd probably wonder how large of a number they can go with, so we can mention this

docs/processors/processor-concurrency.mdx
```diff
@@ -0,0 +1,56 @@
+---
+title: 'Processor Concurrency'
+sidebar_position: 3
+---
+
+Conduit allows you to process records concurrently using the `workers`
```
maha-hajja, line 6:
```suggestion
Conduit allows you to process records concurrently using the processor's `workers`
```

maha-hajja (approved):

PR #38, opened 2023-07-19
Update header to reference new version 0.7

simonl2002:


docusaurus.config.js
```diff
@@ -70,7 +70,7 @@ module.exports = {
           items: [
             {
               label: 'Connectors',
-              to: 'https://github.com/ConduitIO/conduit/blob/main/docs/connectors.md',
+              to: '/docs/connectors/connector-list',
```
hariso, line 73: Good catch!

hariso (approved):
:zap:

PR #41, opened 2023-12-30
Generated connector list

lovromazgon:
This PR adds a generated connector list. The data is taken from a JSON that is also generated by a crawler that finds connectors on GitHub.


simonl2002 (approved):
lgtm

PR #48, opened 2024-03-08
update pipeline config files section

maha-hajja:


docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -276,22 +277,23 @@ beginning.
...
 
 - **Data Type**: String
 - **Required**: Yes
 - **Default**: None
-- **Description**: Defines the processor type (e.g. `js`). Check out the
-  [processors documentation](/docs/processors/getting-started) to find out which
-  processors are supported.
+- **Description**: Defines the processor's plugin name (e.g. `field.set`). Check out the
+  [processors documentation](/docs/processors/builtin) to find the list of builtin processors
+   we provide. Also, check [Building processors](/docs/processors/building-processors)
```
maha-hajja, line 287: note: this document is not created yet :D

raulb: Maybe this could be the link to the page @lovromazgon will create on adding a page about WASM processors?

raulb (commented):
LGTM. I wonder if to not block this PR, we could not add the link to `building processors` and add the reference afterwards once this new page is created.

What do you think? @maha-hajja @lovromazgon

docs/pipeline-configuration-files/specifications.mdx
lovromazgon: Hmm, maybe not the same page, I'm not sure. I thought we'd have a page explaining how to build your own processor and a separate one explaining how WASM processors work underneath the hood (or would that be too in-depth for our users?).

maha-hajja: hmm we could have a page called "WASM in depth" or "processor's architecture" or sth like that, to talk about WASM underneath the hood, but I think this should be separate from the page about building your own processor

raulb: I agree with @maha-hajja . We should keep "how to build" and "in depth/architecture" as separate pages. Maybe add a cross link in between at the bottom to provide additional context. 

Just to clarify, since It seems like we'll need to create an additional page either way, @lovromazgon what did you have in mind to tackle as part of ` Add a page about WASM processors` in https://github.com/ConduitIO/conduit/issues/1422? Would it be "How to build your own processor", maybe?

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -267,7 +268,7 @@ beginning.
 - **Default**: None
 - **Description**: This node contains the configuration of a single processor.
 
-### id
+### plugin
```
lovromazgon, line 271: This should still be id.
```suggestion
### id
```

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -276,22 +277,23 @@ beginning.
   identify the processor inside the pipeline. If multiple processors inside the
   pipeline use the same ID, provisioning will fail.
 
-### type
+### plugin
```
lovromazgon, line 280: We also introduced the field `condition` on a processor.

maha-hajja: that was the reason I opened this PR, how did I forget that 🥲

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -276,22 +278,31 @@ beginning.
...
-  processors are supported.
+- **Description**: Defines the processor's plugin name (e.g. `field.set`). Check out the
+  [processors documentation](/docs/processors/builtin) to find the list of builtin processors
+   we provide.
+
+### condition
+
+- **Data Type**: String
+- **Required**: No
+- **Default**: `True` (all records will passed to the processor)
```
lovromazgon, line 294:
```suggestion
- **Default**: `true` (all records will be passed to the processor)
```

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -276,22 +278,31 @@ beginning.
...
+- **Description**: a Go template expression that will be used as a condition to pass the record to the processor or not.
+the Go template expression will be evaluated, and the evaluated value should either be `True` or `False`, `True` means
+the record will be passed to the processor, `False` means it will be skipped.
```
lovromazgon, line 297: Have you been secretly programming in python⁉️ 🐍 I added some suggestions on clarifying what data us used when evaluating the expression and what happens with "skipped" records (skipped could let somebody think they get filtered out). I also added a link to Go templates docs, as not everybody knows about them.
```suggestion
- **Description**: A [Go template expression](https://pkg.go.dev/text/template) that will be used as a condition to pass the record to the processor or not.
The Go template expression will be evaluated using each record and needs to produce `true` or `false`. If it produces `true`
the record will be passed to the processor, `false` means it will continue down the pipeline without getting processed.
```

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -276,22 +278,31 @@ beginning.
...
 
 ### settings
 
 - **Data Type**: Mapping node
 - **Required**: Yes
 - **Default**: None
 - **Description**: A map of configuration keys and values for the processor. The
-  values in this map depend on the chosen processor [`type`](#type-1).
+  values in this map depend on the chosen processor [`plugin`](#plugin-1).
```
lovromazgon, line 305: Whoops, wrong link
```suggestion
  values in this map depend on the chosen processor [`plugin`](#plugin-2).
```

docs/pipeline-configuration-files/specifications.mdx
maha-hajja: super neat and clear suggestion! thanks

docs/pipeline-configuration-files/specifications.mdx
maha-hajja: nice catch!

docs/pipeline-configuration-files/specifications.mdx
```diff
@@ -43,20 +43,22 @@ pipelines:                      # A list of pipeline configurations
...
           path: ./file2.txt
-        processors:             # A list of processor configurations, processors are attached to the connector
-          - id: proc1           # Processor ID [required]
-            type: js            # Processor type [required]
-            workers: 2          # Number of parallel workers
-            settings:           # A map of configuration keys and values for the processor (specific to the chosen type)
+        processors:                   # A list of processor configurations, processors are attached to the connector
+          - id: proc1                 # Processor ID [required]
+            plugin: custom.javascript # Processor plugin name [required]
+            condition: "{{ eq .Metadata.foo "bar" }}"  # Condition (Go template expression) that dictates if the record will be passed to the processor or not.
```
lovromazgon, line 49: One last thing, since this wouldn't be a valid yaml otherwise:
```suggestion
            condition: '{{ eq .Metadata.foo "bar" }}'  # Condition (Go template expression) that dictates if the record will be passed to the processor or not.
```

lovromazgon (approved):
LGTM!

PR #49, opened 2024-03-08
doc(features): Add page about OpenCDC records

raulb:
## Updates

- [x] Sort featured pages in alphabetical order
- [x] Rename ui.mdx to web-ui.mdx to match page's title.
- [x] Add OpenCDC page 
- [x] Link references in other parts of the conduit documentation about OpenCDC records.

## Meta

Part of https://github.com/ConduitIO/conduit/issues/1422


docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,61 @@
...
+        }
+    }
+}
+```
+
+:::note
+
+Depending on the operation performed on the record, and the connector used its payload can vary. For example:
+
+- When operation is `create`, there won't be a `payload.before`, but a `payload.after` will.
```
maha-hajja, line 49:
```suggestion
- When operation is `create` or `snapshot`, there won't be a `payload.before`, but a `payload.after` will.
```

raulb: Thank you for the suggestion @maha-hajja !

docs/processors/builtin/unwrap.debezium.mdx
```diff
@@ -11,7 +11,7 @@ import Box from "@mui/system/Box";
 
 # `unwrap.debezium`
 
-Unwraps a Debezium record from the input OpenCDC record.
+Unwraps a Debezium record from the input [OpenCDC record](/docs/features/opencdc-record).
```
hariso, line 14: We'll want to change this in the code, i.e. https://github.com/ConduitIO/conduit/blob/main/pkg/plugin/processor/builtin/impl/unwrap/debezium.go#L67

raulb: @hariso you're right. This is autogenerated. I'll take care of that.

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,60 @@
...
+}
+```
+
+:::note
+
+Depending on the operation performed on the record, and the connector used its payload can vary. For example:
+
+- When operation is `create` or `snapshot`, there won't be a `payload.before`, but a `payload.after` will.
+- When operation is `update`, `payload.after` or `payload.before`, any of these could be present.
+- When operation is `delete`, there won't be a `payload.after`, but there will be a `payload.before`sometimes
```
hariso, line 59: A very small thing: space needed before `sometimes` and also a period is missing at the end. Also, maybe it's good to be consistent when it comes to referencing fields and use `.Payload.After` here.

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+## Representation
+
+Conduit relies on [protobuf](https://protobuf.dev/) when it comes to defining an OpenCDC record to benefit from the several advantages that it provides. Its definition can be found in the [Buf Schema Registry](https://buf.build/conduitio/conduit-commons/docs/main:opencdc.v1).
+
+When processing records in Conduit, you can always expect this format,
+
+```json
+{
+    "position": "c3RhbmRpbmc=",
```
hariso, line 30: WDYT about mentioning why we have Base64 here? It might not be obvious to everyone.

docs/processors/builtin/unwrap.debezium.mdx
raulb: @hariso opened this other pull-request to update the source code https://github.com/ConduitIO/conduit/pull/1428.

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,60 @@
...
+    "operation": "update",
+     "metadata": {
```
lovromazgon, line 35:
```suggestion
    "metadata": {
```

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,60 @@
...
+}
+```
+
+:::note
+
+Depending on the operation performed on the record, and the connector used its payload can vary. For example:
+
+- When operation is `create` or `snapshot`, there won't be a `.Payload.Before`, but a `.Payload.After` will.
+- When operation is `update`, `.Payload.After` or `.Payload.Before`, any of these could be present.
+- When operation is `delete`, there won't be a `.Payload.After`, but there will be a `.Payload.Before` sometimes
```
lovromazgon, line 62:
```suggestion
- When operation is `delete`, there won't be a `.Payload.After`, but there will be a `.Payload.Before` sometimes.
```

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,60 @@
...
+* `.Operation` defines what triggered the creation of a record. There are four possibilities: `create`, `update`, `delete` or `snapshot`. The first three operations are encountered during normal CDC operation, while `snapshot` is meant to represent records during an initial load. Depending on the operation, the record will contain either the payload before the change, after the change, both or none (see field `Payload`). 
+* `.Key` represents a value that should identify the entity (e.g. database row).
+* `.Metadata` contains additional information regarding the record.
+* `.Payload` holds the payload change (data before and after the operation occurred).
+
+## Representation
+
+Conduit relies on [protobuf](https://protobuf.dev/) when it comes to defining an OpenCDC record to benefit from the several advantages that it provides. Its definition can be found in the [Buf Schema Registry](https://buf.build/conduitio/conduit-commons/docs/main:opencdc.v1).
+
+When processing records in Conduit, you can always expect this format,
```
lovromazgon, line 29:
```suggestion
When processing records in Conduit, you can always expect this format:
```

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+## Benefits
+
+1. **Support for Operations**: The format should support representing records for create, update, delete, and snapshot operations.
+1. **Standard Metadata Fields**: Definition of standard metadata fields to provide essential information about each record.
+1. **Integration with Data tools**: We believe that by standardizing on this format regardless of the connector type, we'll easily be able to integrate with any other processing data tool that supported this format.
+
+## Fields
+
+* `.Position` uniquely represents the position of record. This is used to track the position of a record in a source enabling Conduit to resume reading from a data stream.
```
lovromazgon, line 19:
```suggestion
* `.Position` uniquely represents the position of record. This is used to track the position of a record in a source connector, enabling Conduit to resume a stopped pipeline.
```

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,60 @@
...
+* `.Operation` defines what triggered the creation of a record. There are four possibilities: `create`, `update`, `delete` or `snapshot`. The first three operations are encountered during normal CDC operation, while `snapshot` is meant to represent records during an initial load. Depending on the operation, the record will contain either the payload before the change, after the change, both or none (see field `Payload`). 
+* `.Key` represents a value that should identify the entity (e.g. database row).
+* `.Metadata` contains additional information regarding the record.
+* `.Payload` holds the payload change (data before and after the operation occurred).
```
lovromazgon, line 23: Maybe we can explain `.Payload.Before` and `.Payload.After` separately?

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,60 @@
+---
+title: 'OpenCDC record'
```
lovromazgon, line 2: I'm missing a word or two about the constants that we provide for some metadata fields, some are even populated by conduit or the connector SDK. The constants are defined [here](https://github.com/ConduitIO/conduit-commons/blob/a6adcbfe1168a50f5f42fdeafa0353eae789dd8e/proto/opencdc/v1/opencdc.proto#L10-L24) (general opencdc constants) and [here](https://github.com/ConduitIO/conduit-connector-protocol/blob/669c83d26b3f9bc58672e69e29a5d94a53af0e2e/proto/connector/v1/connector.proto#L18-L21) (related to conduit connectors). This is for example how we'll add support for multiple tables, to add another constant here that will be used by convention to define where a record was read from or where it should be sent.

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,132 @@
...
+        "opencdc.createdAt": "1663858188836816000",
+        "opencdc.readAt": "1663858188836816000",
+        "opencdc.version": "v1",
+        ...
+    }",
+    ...
+}
+```
+
+### Conduit
```
lovromazgon, line 81: Here we're missing [3 metadata fields](https://github.com/ConduitIO/conduit-commons/blob/a6adcbfe1168a50f5f42fdeafa0353eae789dd8e/opencdc/metadata.go#L59-L67), which are populated when a record is sent to the DLQ. Even though they are not in the proto files (yet), they are important, as they give you insight into why a record landed in the DLQ.

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,132 @@
...
+}
+```
+
+:::note
+`.Position`, `.Key`, and `.Payload.Before` are represented as `Base64` encoded in the example above because these will be a byte slice when represented as JSON.
+:::
+
+## Metadata constants
+
+As part of an OpenCDC record, there will be a set of constants provided for some metadata fields that will vary depending on the connector. These constants can be common to all **OpenCDC** records as part of our standard, some related to our own **Conduit** plugins, and others that will be provided by the **Connector SDK** independently. These constants can be useful to define conventions that will be then used by Conduit to expand its functionality. Here are the ones you can find:
```
lovromazgon, line 60: > some related to our own **Conduit** plugins

Well, not only to plugins, I'd say they are related to Conduit itself, like in general. Like the DLQ metadata fields I mentioned in the other comment, those are not related to a Conduit plugin.

> others that will be provided by the **Connector SDK** independently

The last group are metadata fields provided by connector implementations. Each connector is allowed to add any metadata it wants, the SDK doesn't have anything to do with it.

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,132 @@
...
+        "conduit.destination.plugin.version": "v1",
+        ...
+    }",
+    ...
+}
+```
+
+### Connector SDK
+
+These metadata fields will depend on the Connector SDK, but its contants will use the same namespace format as above by convention, indicating first the connector name that are providing them.
```
lovromazgon, line 119: This also needs to be reworded based on what I wrote in the other comment about custom metadata provided by actual connector implementations.

Oh, and we could mention that the convention of adding the connector name as a prefix is to prevent accidental name clashes of metadata keys. This way we're sure that `postgres.table` for instance is populated only by a `postgres` connector and no other.

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,132 @@
...
+
+Using the same [example above](#representation), you'll find there is a metadata key named `file.path`, and this indicates a metadata constant named `file.path` provided by the `file` builtin plugin.
```
lovromazgon, line 121: Scratch builtin, it could be a standalone plugin. Also I wouldn't call this a constant, it's just a value that gets added by the connector to metadata, who knows if it's a constant or not.

```suggestion
Using the same [example above](#representation), you'll find there is a metadata key named `file.path`, which indicates a metadata value provided by the `file` plugin.
```

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,132 @@
...
+
+An OpenCDC record in Conduit aims to standardize the format of data records exchanged between different connectors within a data processing pipeline. The primary objective is to ensure compatibility between various combinations of source and destination connectors.
+
+## Benefits
+
+1. **Support for Operations**: The format should support representing records for create, update, delete, and snapshot operations.
+2. **Standard Metadata Fields**: Definition of standard metadata fields to provide essential information about each record. These can vary depending on the record. See [Metadata Constants](#metadata-constants) for more information.
+3. **Integration with Data Tools**: We believe that being strict about the record format Conduit consumes and produces will make it easier to integrate with other data processing tools.
+
+## Fields
```
lovromazgon, line 14: Can we add a `:::note` callout at the end of this chapter explaining why we are writing `.Position` and not `.position`? Basically explain that we are using the [Go templates](https://pkg.go.dev/text/template) notation for the Go representation of the record (we could even link to it), and because the fields are public they start with an uppercase letter.

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,143 @@
...
+    "metadata": "{
+        "conduit.destination.plugin.name": "example",
+        "conduit.destination.plugin.version": "v1",
+        ...
+    }",
+    ...
+}
+```
+
+When a record is sent to the [Dead-Letter Queue (DLQ)](/dead-letter-queue), you'll also see these extra fields that will give you an insight into why the landed in the DQL.
```
lovromazgon, line 123: You need to stop dropping records, this is breaking our record delivery guarantees!
```suggestion
When a record is sent to the [Dead-Letter Queue (DLQ)](/dead-letter-queue), you'll also see these extra fields that will give you an insight into why the record landed in the DLQ.
```

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,143 @@
...
+
+- `conduit.dlq.nack.error` contains the error that caused a record to be nacked and pushed to the dead-letter queue.
+- `conduit.dlq.nack.node.id` is the ID of the internal node that nacked the record. 
+
+### Connector
+
+These metadata fields will be provided by each connector implementation allowing them to add any necessary metadata. As previously mentioned, to avoid unintended conflicts of metadata keys, the convention these will follow are the same as before, indicating first the connector name that's adding them.
+
+Taking the same [previous record example](#representation), you'll notice there is a metadata key named `file.path`, which would indicate this is record processed by a `file` plugin.
```
lovromazgon, line 132:
```suggestion
Taking the same [previous record example](#representation), you'll notice there is a metadata key named `file.path`, which would indicate this field was added by a `file` plugin.
```

docs/features/opencdc-record.mdx
```diff
@@ -0,0 +1,143 @@
...
+
+```json
+{
+    "position": "c3RhbmRpbmc=",
+    "operation": "update",
+    "metadata": "{
+        "file.path": "./example.in",
+        "opencdc.readAt": "1663858188836816000",
+        "opencdc.version": "v1"
+    }",
```
lovromazgon, line 41: Metadata is a map, not a string. We need to update the other examples as well.
```suggestion
    "metadata": {
        "file.path": "./example.in",
        "opencdc.readAt": "1663858188836816000",
        "opencdc.version": "v1"
    },
```

lovromazgon (approved):
🥇

PR #50, opened 2024-03-08
Referencing fields

hariso:
Part of https://github.com/ConduitIO/conduit/issues/1422, includes specification changes from https://github.com/ConduitIO/conduit/pull/1424.


raulb (approved):
Nice! @hariso Great call already pointing out to the page that will be added as part of https://github.com/ConduitIO/conduit-site/pull/49. Will make sure I ship that one soon so we can ship this one without a broken link.

docs/processors/referencing-fields.mdx
```diff
@@ -0,0 +1,47 @@
+---
+title: 'Referencing record fields'
+sidebar_position: 1
+---
+
+Very often, processors can be configured to work on a specific field in a
+record (which is always in
+the [OpenCDC format](/docs/features/opencdc-record.mdx)). That is done through
```
raulb, line 8: I just realized about this.

```suggestion
the [OpenCDC format](/docs/features/opencdc-record)). That is done through
```

raulb (changes requested):
I just realized that we need to link differently to this page.

docs/processors/referencing-fields.mdx
```diff
@@ -0,0 +1,47 @@
+---
+title: 'Referencing record fields'
+sidebar_position: 1
+---
+
+Very often, processors can be configured to work on a specific field in a
```
maha-hajja, line 6: I think we should also mention that these references are how the fields would be addressed in a Go template, and maybe add a link for go templates format

raulb (approved):
Looks great. I really like having both the json and the dot notation side by side, extending this information below. Great work 👌

PR #51, opened 2024-03-08
Add a 'Referencing processors' page

hariso:
Shamelessly taken from the existing page about referencing connectors.:)


docs/processors/referencing.mdx
```diff
@@ -0,0 +1,44 @@
...
+      if the standalone plugin wasn't found
+- `builtin:base64.decode`
+    - will use the **latest** **builtin** **base64.decode** plugin
+- `standalone:base64.decode@v0.3.0`
+    - will use the **standalone** **base64.decode** plugin with version **v0.3.0** 
+      (no fallback to builtin)
+
+:::info
+To read more about how to register standalone processors in Conduit, please
+refer to [WASM processors](/docs/processors/wasm-processors.mdx).
```
raulb, line 43:
```suggestion
refer to [WASM processors](/docs/processors/wasm-processors).
```

raulb: @lovromazgon to confirm this will be the correct link.

lovromazgon: Not exactly sure yet, my first instinct is to put the page under `docs/processors/standalone-processors/wasm`, since WASM is only relevant for standalone processors. Regardless, it should be without `.mdx` (as per your suggestion above). We can clean up broken links at the end, docusaurus warns us about them, so these are easy to find:

```
[WARNING] Docs markdown link couldn't be resolved: (/docs/processors/wasm-processors.mdx) in "/Users/lovro/dev/conduitio/conduit-site/docs/processors/referencing.mdx" for version current
```

lovromazgon: Actually, now that I read the sentence again, this seems to be a link that should point to `/docs/processors/standalone-processors`.

PR #53, opened 2024-03-11
add "building standalone processors" document

maha-hajja:
this doc is pretty much the same as the processor SDK document.
do we need to add any more details?

Part of https://github.com/ConduitIO/conduit/issues/1422


raulb (approved):
@maha-hajja Page looks good. I only have a question. Do you think in the future we could have page dedicated to another language? This page looks how to build a standalone processor using Go, so I wonder if we should already start making that distinction somewhere. What do you think?

lovromazgon:
Good point @raulb, we might want to already start preparing separate docs and explain somewhere that we currently only provide a Go SDK, but that we plan to add more languages in the future, and call out that people can open an issue to request a specific one.

I'm also thinking that we could simplify the processor SDK readme. We can only show the simple example using `sdk.NewProcessorFunc` and then link to this page for more complex processors. This way we wouldn't scare away people when they open the processor SDK. WDYT?

maha-hajja:
@lovromazgon yes I like that idea! this way this documentation won't be a duplicate of the readme, and as you said it won't scare them away.
about using other languages for processors, I think this page is a good place to mention that fact and link to the issues, but I don't think we need to add more pages for the sake of other languages now. WDYT @raulb

raulb:
@maha-hajja @lovromazgon good idea on adding a link there to create an issue requesting a new language. I agree that we probably don't need to worry about structuring the page based on new languages for now.

docs/processors/building-standalone-processors.mdx
```diff
@@ -0,0 +1,135 @@
+---
+title: 'Building Standalone Processors'
```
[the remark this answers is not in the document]
lovromazgon: I'd go a different route and do something like this:
```
processors
├── ... # Other pages
└── standalone
    ├── index.mdx    # General info
    │                # about how to use
    │                # standalone processors.
    ├── building.mdx # This page - how
    │                # to build your own
    │                # standalone processor.
    └── wasm.mdx     # The page I'm working
                     # on - how standalone
                     # processors work under
                     # the hood.
```

In the menu it would then look something like:
```
Standalone processors >
  How to build a standalone processor
  How it works (WASM)
```

The naming of files and pages can be adjusted, I'm mostly proposing the nesting structure.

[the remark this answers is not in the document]
maha-hajja: I like the nesting structure too

docs/processors/standalone/building.mdx
```diff
@@ -0,0 +1,141 @@
+---
+title: 'Build your own'
```
maha-hajja, line 2: I was on edge between `Building` and `Build your own`, WDYT? @lovromazgon @raulb

docs/processors/building-standalone-processors.mdx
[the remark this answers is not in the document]
maha-hajja: done

docs/processors/standalone/building.mdx
raulb: @maha-hajja my proposal is (also mainly for your SEO purposes) to go with `How to build your own`.

raulb, line 2: We could also rename the file to `how-to-build.mdx`

```suggestion
title: 'How to build your own standalone processor'
```

raulb: Actually, if we go with the route of "How to", it's quite possible that most of our pages become "how to..." which could turn counterproductive. 

@maha-hajja If we go with how other pages are titled, maybe it should be "Building".

lovromazgon: I don't have strong opinions here, I think `Build your own` is not bad though 🤷

docs/processors/standalone/building.mdx
```diff
@@ -0,0 +1,141 @@
...
+sidebar_position: 0
```
lovromazgon, line 3: Sorry for this long comment in advance, and I should have probably posted it sooner, but the more I think about it, the more I see the need to restructure this page entirely. I'm posting this comment on this line, simply so that we can have a discussion.

My proposal is to start from scratch and not follow the readme we had in the SDK. I see this as more of an in-depth page that explains in detail what to do, while the readme previously only tried to give a quick overview of the most important parts. I sketched out the structure I imagined:

---

# Build your own

Explain that you can build your own processor using the SDK, that we currently provide a Go SDK, that you may poke us in an issue if you want a different language, that you can look at the Wasm page if you want to build your own.

Mention that the SDK exposes two ways of building a processor, one for very simple ones, the other that is a bit more complicated, but gives you more control.

## Using `sdk.NewProcesorFunc`

Explain `sdk.NewProcessorFunc` approach. Explain when to use it, what are its limitations and when to switch to the full processor approach.

## Using `sdk.Processor`

Explains how to build the full-blown processor. I'd try to break this down into the parts that need to be implemented.

### Specifications

Explain what are specifications, what are they good for. Definitely mention `paramgen` and explain how to use it to generate the parameter specifications.

### Configure

Explain what to implement (and what _not_ to) in this function. Definitely mention the utilities we provide and how most configuration implementations will look like. Mention how to add custom validations here.

### Open

Explain what is supposed to happen here (e.g. opening connections, initializing resources). Mention this one is optional.

### Process

Here we should go into detail on what to take into account when implementing this.

### Teardown

Explain that this is the counterpart to `Open`, that it's meant to close any open connections and resources. This one is optional, same as `Open`.

## Logging

Explain that the SDK provides a logger, that it should be retrieved using the context provided by Conduit, that log level should be "trace" in the hot path.

## Compiling the processor

Mention the env variables that need to be set and why that's the case.

At the end point to the main "Standalone processors" page to see how the compiled processor can then be used in a pipeline.

---

Let me know what you think. cc @raulb @hariso

raulb: @lovromazgon I agree with that suggestion. I see the value of presenting different sets of information between the repository readme and this page. I think we should also go in the direction of showing code snippets of more in depth language instructions on each GitHub repository, and leaving these documents to express more context on what it is, and why it could be useful.

I like the structure of what you're presenting because it reminds me of what we have in other pages recently added / updated which can give you a quick overview of what's possible (or not -suggesting to create an issue-), and then exposing the different options and _components_ with a different language.

raulb (approved):
Looks great! :shipit:

PR #57, opened 2024-03-13
Standalone processors under the hood (WASM)

lovromazgon:

### DIFF
--- a/docs/processors/standalone/wasm.mdx
+++ b/docs/processors/standalone/wasm.mdx
@@ -51,8 +51,8 @@ sequenceDiagram
    This call is blocking and will return when a command is available.
 4. Conduit populates the allocated memory with a command and returns the size of
    the command or an error code. See [error codes](#error-codes) for more a list
-   of possible error codes. See [Commands](#commands)
-   for more information about the command format.
+   of possible error codes. See [`command_request`](#command_request) for more
+   information about the command format.
 5. In case Conduit returns an error code, the processor should exit with the
    error code.
 6. The processor parses and executes the command. Note that everything except
@@ -60,7 +60,7 @@ sequenceDiagram
    processor SDK.
 7. The processor calls `command_response` to send the result back to Conduit.
    This call should happen even if the command execution failed (the result
-   should include the error). See [Commands](#commands)
+   should include the error). See [`command_response`](#command_response)
    for more information about the response format.
 8. Conduit acknowledges the result and the processor can continue retrieving
    the next command.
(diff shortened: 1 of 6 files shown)


docs/processors/standalone/wasm.mdx
```diff
@@ -0,0 +1,229 @@
+---
+sidebar_position: 2
+---
+
+# How it works under the hood (Wasm)
```
raulb, line 5: I'm wondering if we should simply could call it "How it works" and then explain in the document why using Wasm in the first place.

I think that could still valuable to explain the versatility this brings.

raulb (approved):
Looks great. I love seeing these going more in depth. I only have one question/comment on maybe adding an explanation of why wasm, or the benefits this brings us while also maybe tweaking the title of this page.

@lovromazgon what do you think?

docs/processors/standalone/wasm.mdx
lovromazgon: I was expecting a comment about the title, I'm not too happy with it myself. `How it works` sounds better, the `under the hood` part is kinda redundant. I'm wondering if there's a way to include `Wasm` in the title without making it too long? 🤔 I feel like with that in the title you already get an idea that standalone processors are Wasm modules, without even opening the page.

I'll add an explanation on top why Wasm is a good choice 👍

lovromazgon: Done!

docs/processors/standalone/wasm.mdx
```diff
@@ -0,0 +1,259 @@
+---
+sidebar_position: 2
+---
+
+# How it works
+
+Here we will describe how Conduit runs standalone processors under the hood.
+This is useful to understand when writing a processor SDK.
+
+## Why WebAssembly?
```
raulb, line 10: Making this suggestion to be consistent with we express these in our docs (e.g.: "How it works" vs "How does it work?")

```suggestion
## Why WebAssembly
```

raulb (approved):
Approved with just a minor suggestion. I really liked the introduction about why we chose to use wasm for our processors, and the options these provide. Nice work.

docs/processors/standalone/wasm.mdx
```diff
@@ -0,0 +1,259 @@
+---
```
raulb, line 1: @lovromazgon I just realized it'd be probably better if we renamed this page to `how-it-works.mdx`

docs/processors/standalone/wasm.mdx
lovromazgon: I simplified this header and renamed it to `WebAssembly` 😉

PR #69, opened 2024-04-02
Automatically rebuild connectors list

### LINKED ISSUE #66: Automatically update list of connectors
The list of connectors on our website ([link](https://conduit.io/docs/connectors/connector-list/)) is taken from [connectors.json](https://github.com/ConduitIO/conduit-site/blob/main/static/connectors.json) that is manually built using [a script](https://github.com/lovromazgon/connector-list).

We should automate this process and let it run periodically. The expected output should be a pull request, that is merged if it's deemed "safe" (definition of "safe" TBD but generally it's about preventing unintentional updates).

hariso:
This PR adds `connectorgen`, a tool that rebuilds the list of connectors, and a GitHub workflow that invokes `connectorgen` once a week.

`connectorgen` is the Go version of a great script by @lovromazgon that can be found here: https://github.com/lovromazgon/connector-list/.

I needed to make a few small adjustments to the script (sorting repos by URL), so I thought that this change and future changes in the logic can be more easily done in Go. Also, the used Go client is configured to automatically apply rate limiting. In future, it should be easier to use tool to compare the existing `connectors.json` with the generated one so that we can determine if the changes are generally safe to merge.

The PR also includes initial changes to `connectors.json`. 

Resolves #66.


.github/workflows/update-connectors-list.yaml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Update connectors' list
+        working-directory: src/connectorgen
+        run: |
+          go run main.go ${GITHUB_WORKSPACE}/static/connectors.json
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v6.0.2
```
lovromazgon, line 22: Can we also auto-merge the PR? No need to have the safety check (that's future work), but auto merging would be nice to keep the manual toil to a minimum, otherwise we'll soon be annoyed by this.

I didn't look too far into this, but I see a similar action that should allow us to do that [peter-evans/enable-pull-request-automerge](https://github.com/peter-evans/enable-pull-request-automerge).

hariso: Hmm, I guess that's fine... After all, we can easily rebuild the list or revert the change (if needed).

hariso: Done!

raulb (approved):
😍 Thank you @hariso !!

lovromazgon (approved):

PR #70, opened 2024-04-03
doc: update conduit release

raulb:


docusaurus.config.ts
```diff
@@ -132,7 +132,7 @@ const config: Config = {
     },
     announcementBar: {
       id: 'announcement-bar-2', // increment on change
```
lovromazgon, line 134: Sidenote: If we want to show the announcement bar to everyone who already dismissed it, we need to increment this id. That said, Given that it's a minor release, I don't think we necessarily need to do it here.

lovromazgon (approved):

docusaurus.config.ts
raulb: @lovromazgon TIL!

PR #73, opened 2024-04-03
Display processor configuration as YAML

lovromazgon:
This makes it possible to display the processor configuration as YAML in addition to the table we display now.

### DIFF
--- a/src/processorgen/mdx.tmpl
+++ b/src/processorgen/mdx.tmpl
@@ -8,6 +8,8 @@ sidebar_position: {{ .index }}
 import ReactDiffViewer from 'react-diff-viewer';
 import Chip from '@mui/material/Chip';
 import Box from "@mui/system/Box";
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
 
 # `{{ .specification.name }}`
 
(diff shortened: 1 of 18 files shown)


src/processorgen/mdx.tmpl
```diff
@@ -19,28 +21,52 @@ import Box from "@mui/system/Box";
...
-      <td>{{ $param.type }}</td>
-      <td>{{ formatParameterValue $param.default }}</td>
-      <td>
-{{ $param.description }}
-      </td>
-    </tr>
-    {{- end }}
-  </table>
-{{- end }}
+<Tabs groupId="config-params">
```
hariso, line 24: WDYT about leaving just the YAML? It's very easy to use to look up parameters (which is what we'd do with the table)?

hariso (approved):
:rocket:

src/processorgen/mdx.tmpl
lovromazgon: If we want to only show yaml we should somehow also display the parameter type (for example as an inline comment).

Another way would be to flip this around and have YAML as the default and still allow the user to switch to the table view.

hariso: At least to me, the first option is more appealing, because I can more easily work with the YAML. The 2nd option is good too.:)

raulb (commented):
✋ 
🎤

raulb (approved):

maha-hajja (approved):
nice addition!

PR #79, opened 2024-04-19
Record routing

hariso:
Closes https://github.com/ConduitIO/conduit/issues/1493.

As suggested in the issue, the page is named "Record routing". Routing the records with standard `opencdc.collection` metadata is possible in some scenarios, but in others, when we can use other features in Conduit.

### DIFF
--- a/docs/features/record-routing.mdx
+++ b/docs/features/record-routing.mdx
@@ -18,15 +18,21 @@ Through this, a number of use cases are possible:
   multiple tables in each can be kept in sync with a single pipeline with only
   one source and one destination connector.
 * Route data from one collection (table, topic, index, etc.) to multiple
-  collections.
+  collections on a single server.
+* Route data to different instances or different systems altogether.
 
-## How does it work
+This can be done it two ways:
+
+* by using a connector that supports writing to multiple collections
+* by using filter processors
+
+## Using a connector that supports writing to multiple collections
 
 The [OpenCDC format](/docs/features/opencdc-record) recommends that a record's
 source or destination collection is written to its metadata using
 the [`opencdc.collection` key](/docs/features/opencdc-record#opencdccollection).
 
-Then, a destination connector can utilize that information and write the record
+A destination connector can then utilize that information and write the record
 to the correct collection.
 
 :::note
@@ -35,10 +41,12 @@ source collections, or write to multiple destination collections. Consult the
 connector documentation to check if it supports this feature.
 :::
 
-## Example
+### Example: Route data from multiple Kafka topics to multiple PostgreSQL tables on a single database server
 
 Let's assume we have a pipeline that streams data from two Kafka
-topics, `employees_finance` and `employees_legal` to PostgreSQL.
+topics, `employees_finance` and `employees_legal` to PostgreSQL (both of the
+connectors support reads from multiple collections and writes to multiple
+collections). A pipeline configuration file could look like this:
 
 ```yaml
 ---
@@ -66,7 +74,7 @@ pipelines:
           table: "employees"
 ```
 
-The Kafka connector will connectors like below:
+The Kafka connector will read records like the ones below:
 
 ```json lines
 {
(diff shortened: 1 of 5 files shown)


docs/features/record-routing.mdx
```diff
@@ -0,0 +1,177 @@
...
+}
+```
+
+The PostgreSQL destination connector will check the value in
+the `opencdc.collection` metadata field. Based on that, it will write John's
+information to the `employees_finance` table and Adam's information to
+the `employees_legal` table.
+
+If the `opencdc.collection` metadata field is missing, the record will then be
+written to the default table, which is `employees` in our case.
```
lovromazgon, line 112: This is not correct. This was the old behavior, where the connector looked at the metadata field `postgres.table` and used the connector configuration parameter `table` as a fallback, if that metadata field was not found.

Now we leave it to the user to configure how they want the connector to figure out the target table. The configuration parameter `table` accepts a Go template, which will be evaluated for each record, the resulting string will be used as the target table. By default, the configuration parameter is set to `{{ index .Metadata "opencdc.collection" }}`, meaning that by default we try to route the record based on the metadata field `opencdc.collection`. However, if you set `table` to `employees`, then all records will be written into `employees`, including ones that contain the aforementioned metadata field.

docs/features/record-routing.mdx
```diff
@@ -0,0 +1,177 @@
...
+
+      - id: finance
+        type: destination
+        plugin: "builtin:postgres"
+        name: finance
+        settings:
+          url: "postgresql://username:password@localhost/finance-db?sslmode=disable"
+          table: "employees"
+        processors:
+        - id: employees_finance_only
```
lovromazgon, line 162: Nitpick: indentation
```suggestion
          - id: employees_finance_only
```

docs/features/record-routing.mdx
```diff
@@ -0,0 +1,177 @@
...
+          plugin: "filter"
+          condition: `{{index .Metadata "department" | eq "finance" | not}}`
```
lovromazgon, line 164: I'd add a comment above this line, explaining what the condition is doing, something like:
```suggestion
          # Filter out records, which don't contain the value "finance" in the metadata field "department"
          condition: `{{index .Metadata "department" | eq "finance" | not}}`
```

docs/features/record-routing.mdx
```diff
@@ -0,0 +1,177 @@
...
+        type: destination
+        plugin: "builtin:postgres"
+        name: legal
+        settings:
+          url: "postgresql://username:password@localhost/legal-db?sslmode=disable"
+          table: "employees"
+        processors:
+        - id: employees_legal_only
+          plugin: "filter"
+          condition: `{{index .Metadata "department" | eq "legal" | not}}`
```
lovromazgon, line 176: Btw, since `department` is a valid identifier, we could take the shortcut and use `.Metadata.department`. ~I think it should work~ [I tried it out](https://go.dev/play/p/j6VLIS5neqY).
```suggestion
          condition: `{{ eq .Metadata.department "legal" | not }}`
```

docs/features/record-routing.mdx
```diff
@@ -0,0 +1,177 @@
...
+  multiple tables in each can be kept in sync with a single pipeline with only
+  one source and one destination connector.
+* Route data from one collection (table, topic, index, etc.) to multiple
+  collections on a single server.
+* Route data to different instances or different systems altogether.
+
+This can be done it two ways:
+
+* by using a connector that supports writing to multiple collections
+* by using filter processors
```
lovromazgon, line 27:
```suggestion
* using a connector that supports writing to multiple collections,
* using filter processors.
```

docs/features/record-routing.mdx
```diff
@@ -0,0 +1,177 @@
...
+Through this, a number of use cases are possible:
+
+* Keep two instances synchronized. For examples, two database instances with
+  multiple tables in each can be kept in sync with a single pipeline with only
+  one source and one destination connector.
+* Route data from one collection (table, topic, index, etc.) to multiple
+  collections on a single server.
+* Route data to different instances or different systems altogether.
+
+This can be done it two ways:
```
lovromazgon, line 24:
```suggestion
This can be done in two ways:
```

docs/features/record-routing.mdx
hariso: Ah, I see, so no fallback.

docs/features/record-routing.mdx
hariso: Done in [51cba2e](https://github.com/ConduitIO/conduit-site/pull/79/commits/51cba2e59fe619c036c820a3d91b7e7f4b44fedd).

docs/features/record-routing.mdx
```diff
@@ -0,0 +1,175 @@
...
+     
+      - id: legal
+        type: destination
+        plugin: "builtin:postgres"
+        name: legal
+        settings:
+          url: "postgresql://username:password@localhost/legal-db?sslmode=disable"
+          table: "employees"
+        processors:
+        - id: employees_legal_only
```
lovromazgon, line 172: Another indentation nitpick 🙃 
```suggestion
          - id: employees_legal_only
```

docs/features/record-routing.mdx
```diff
@@ -0,0 +1,175 @@
...
+        plugin: "builtin:postgres"
+        name: finance
+        settings:
+          url: "postgresql://username:password@localhost/finance-db?sslmode=disable"
+          table: "employees"
+        processors:
+          - id: employees_finance_only
+            plugin: "filter"
+            # Filter out records where the metadata field "department" is NOT set to "finance".
+            condition: `{{index .Metadata.department | eq "finance" | not}}`
```
lovromazgon, line 162: This needs to be fixed
```suggestion
            condition: `{{ eq .Metadata.department "finance" | not }}`
```

lovromazgon (approved):
LGTM, a few more nitpicks, but approving in advance 👍

docs/features/record-routing.mdx
lovromazgon: This one slipped through the cracks @hariso ☝️

hariso:
Thanks for all the suggestions @lovromazgon !

PR #87, opened 2024-05-07
add scarf pixels for each section

maha-hajja:
created and added a new scarf pixel for each section of the docs.

Fixes https://github.com/ConduitIO/conduit/issues/1535

### DIFF
--- a/docs/connectors/connector-list.mdx
+++ b/docs/connectors/connector-list.mdx
@@ -2,7 +2,7 @@
 title: "Connector List"
 sidebar_position: 2
 ---
-
+![scarf pixel](https://static.scarf.sh/a.png?x-pxid=2fa824d7-fd94-4cf9-a5c8-ea63c9860213)
 The Conduit team and our community of developers are always adding new connectors.
 Don't see the connector you need below? Help us prioritize by creating new issues,
 commenting, or leaving a `+1` on our [source](https://github.com/ConduitIO/conduit/issues?q=is%3Aissue+label%3Aconnector%3Asource+is%3Aopen)
--- a/docs/connectors/connector-list.mdx
+++ b/docs/connectors/connector-list.mdx
@@ -9,7 +9,7 @@ commenting, or leaving a `+1` on our [source](https://github.com/ConduitIO/condu
 and [destination](https://github.com/ConduitIO/conduit/issues?q=is%3Aissue+label%3Aconnector%3Adestination+is%3Aopen)
 connector lists.
 
-
+![scarf pixel](https://static.scarf.sh/a.png?x-pxid=2fa824d7-fd94-4cf9-a5c8-ea63c9860213)
 Don't have time to wait? You can get started [building your own](/docs/connectors/building) in no time.
 
 ### Built-in vs Standalone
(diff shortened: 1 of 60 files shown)


src/pages/index.tsx
```diff
@@ -19,51 +19,55 @@ import styles from './index.module.css';
...
-                </a>
-
-                <a href="/docs/introduction/getting-started"
-                   className={clsx(styles['bg-transparent-200'], "inline-flex items-center px-4 py-2 border border-transparent shadow-sm text-sm font-medium rounded-md text-white hover:text-gray-600 hover:bg-white focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-indigo-500")}
-                >
-                  <span>Documentation</span>
-                </a>
+      <>
+        <img referrerPolicy="no-referrer-when-downgrade"
+             src="https://static.scarf.sh/a.png?x-pxid=80cfd08d-1f82-4068-960a-cee31e1ae3fa"/>
```
maha-hajja, line 24: note that this is the only change on this page.. just to make it easier for reviews

docs/introduction/vocabulary.mdx
```diff
@@ -2,7 +2,7 @@
 title: "Vocabulary"
 sidebar_position: 3
 ---
-
+![scarf pixel](https://static.scarf.sh/a.png?x-pxid=01346572-0d57-4df3-8399-1425db913a0a)
```
lovromazgon, line 5: I don't like how this pixel adds a whole empty line to our pages, for instance:

image

Is there a way to make the pixel actually invisible? Otherwise I propose to move the pixel to the bottom of all pages, the empty line at the end won't be as noticable.

src/pages/index.tsx
```diff
@@ -19,51 +19,54 @@ import styles from './index.module.css';
...
-                <a href="/docs/introduction/getting-started"
-                   className={clsx(styles['bg-transparent-200'], "inline-flex items-center px-4 py-2 border border-transparent shadow-sm text-sm font-medium rounded-md text-white hover:text-gray-600 hover:bg-white focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-indigo-500")}
-                >
-                  <span>Documentation</span>
-                </a>
+      <>
+        <div style={{'background': 'linear-gradient(289.62deg, #20BED9 -11.02%, #0B525D 97.11%)'}}
+             className="max-w-full">
+          <img referrerPolicy="no-referrer-when-downgrade"
+               src="https://static.scarf.sh/a.png?x-pxid=80cfd08d-1f82-4068-960a-cee31e1ae3fa"/>
```
lovromazgon, line 26: This one also pushes the navbar down quite a bit. I'd either style the image so it's 0 (or 1) pixel in height, or add it to the footer instead.

docs/connectors/connector-list.mdx
```diff
@@ -2,7 +2,7 @@
 title: "Connector List"
```
lovromazgon, line 2: On the connector list, we should change the release download links to go through scarf. Might be a thing for a separate PR though.

maha-hajja: yes that's on my list.. trying to figure out where we set the URL values so I could update them for the json list, would appreciate some quick pointers since you created that generator 👀

src/pages/index.tsx
maha-hajja: even in the footer it made a mess, I added it after a paragraph at the bottom of the page so it doesn't affect anything else.

raulb (approved):
@maha-hajja LGTM. Something I would personally find useful or easier to maintain long term is to use some description so we know to what section / pages each pixel relates to.

e.g.: `connectors scarf pixel`

Just an idea!

PR #88, opened 2024-05-09
feat: site updates

raulb:
- [x] Upgrade dependencies
- [x] Consolidate `/docs` as the main url for _Documentation_ to apply to Algolia Search with a better URL.
- [x] Update Twitter logo

**Before**

image

**After**

image

- [x] Restructure pages consolidating what's duplicated


docs/getting-started/installing-and-running.mdx
```diff
@@ -0,0 +1,144 @@
...
+$ make build-server
+```
+
+You will have a new binary built for your architecture and machine. All that's left is to run it!
+
+```shell
+./conduit
+```
+  </TabItem>
+  <TabItem value="homebrew" label="Homebrew">
```
lovromazgon, line 103: IMO homebrew should be the first provided option.

docs/getting-started/installing-and-running.mdx
```diff
@@ -0,0 +1,144 @@
...
+2024-02-20T21:37:45+00:00 INF http server started address=[::]:8080
+2024-02-20T21:37:45+00:00 INF
+2024-02-20T21:37:45+00:00 INF click here to navigate to Conduit UI: http://localhost:8080/ui
+2024-02-20T21:37:45+00:00 INF click here to navigate to explore the HTTP API: http://localhost:8080/openapi
+2024-02-20T21:37:45+00:00 INF
+```
+
+You should now be able to interact with the Conduit UI and HTTP API on port 8080.
+  </TabItem>
+  <TabItem value="source" label="Building from source">
```
lovromazgon, line 64: While building from source should be last.

docs/introduction.mdx
```diff
@@ -20,25 +20,39 @@ and [any datastore you can build a plugin for](/docs/connectors/building).
...
-chmod a+x conduit
-```
+Conduit connectors are plugins that communicate with Conduit via a gRPC
+interface. This means that plugins can be written in any language as long as
+they conform to the required interface. Check out
+our [connector docs](/docs/connectors/installing)!
+
+## Installing
+
+To get started make sure you download the latest version of Conduit using one of our available options.
```
lovromazgon, line 53: I would provide a snippet how to install it via homebrew and then refer to the install page for other methods, otherwise the reader needs to navigate to a different page, which kind of breaks the flow.

src/components/HomepageNavbar/index.tsx
```diff
@@ -9,7 +9,7 @@ const navLinks = [
   },
   {
     label: 'Documentation',
-    href: '/docs/introduction/getting-started',
```
lovromazgon, line 12: We have to update the link in the Conduit readme as well, maybe there's other places.

docs/getting-started/getting-started.mdx
```diff
@@ -0,0 +1,11 @@
...
+title: 'Getting Started with Conduit'
+sidebar_position: 0
+sidebar_label: "Getting Started"
+---
+
+import DocCardList from '@theme/DocCardList';
+
+To help you get started, here are some of the most popular pages in our documentation:
+
+<DocCardList/>
```
lovromazgon, line 11: One thing that bothers me a bit is that some expandable pages are clickable, while others aren't (I mean the pages that contain nested pages). I would love it if we could consolidate that, so it's consistent. WDYT?

docs/getting-started/installing-and-running.mdx
```diff
@@ -0,0 +1,144 @@
...
+A file called `conduit.tgz` will be in your current directory. The next step is to expand the source:
+
+```shell
+$ tar zxvf conduit.tgz
+```
+
+Then change directories to the appropriate folder. Keep in mind that the folder name might be different between releases since it's tied to the latest git sha for the commit.
+
+```shell
+$ cd ConduitIO-
```
lovromazgon, line 82: This looks a bit strange, I would specify the whole folder name, even though it will change in future releases. I think people will get what we mean.
```suggestion
$ cd ConduitIO-conduit-0ce62ec
```

docs/getting-started/installing-and-running.mdx
```diff
@@ -0,0 +1,144 @@
...
+
+You should now be able to interact with the Conduit UI and HTTP API on port 8080.
+  </TabItem>
+  <TabItem value="source" label="Building from source">
+Before you can build Conduit from source, you need to have Go 1.21 or higher installed. The Go website has [comprehensive instructions](https://go.dev/doc/install) on how to get the language installed on your machine. Once installed, you can follow the rest of the instructions.
+
+1. Start by downloading the source code from the latest stable release on the Conduit [Releases Page](https://github.com/ConduitIO/conduit/releases/latest). Alternatively, you can run this command to automatically download the latest stable source to your current directory:
+
+```shell
+$ TAG=$(curl -s https://api.github.com/repos/ConduitIO/conduit/releases/latest | grep "tag_name\": \"v[0-9.]*" | awk -F \" '/tag_name\": \"*\"/{print $(NF-1)}') && curl -o conduit.tgz -L https://github.com/ConduitIO/conduit/tarball/$TAG
```
lovromazgon, line 70: I propose to hardcode the tag here and update it whenever we release (we already have other places in this doc that we need to update, so I don't see much harm in that). Additionally, I would use the URL `https://github.com/ConduitIO/conduit/archive/refs/tags/v0.10.0.tar.gz`, since that tarball contains a folder named `conduit-0.10.0` which then makes it simpler to describe how to `cd` into that folder, as the path won't contain a git SHA in the name.

```suggestion
$ TAG=v0.10.0 curl -o conduit.tar.gz -L https://github.com/ConduitIO/conduit/archive/refs/tags/$TAG.tar.gz
```

lovromazgon (commented):
Great changes @raulb 👏 I added a few suggestions, lmk what you think.

docs/getting-started/getting-started.mdx
raulb: @lovromazgon I agree. That's also something that I wanted to change. I was considering only changing the pages I touched, but I can squeeze that change in as well.

src/components/HomepageNavbar/index.tsx
raulb: @lovromazgon I'll update the internal links and will add a redirect in case that's linked from elsewhere.

raulb:
@lovromazgon thanks for the feedback. I think I addressed it all. Let me know what you think. There's still something I'd like to do (if possible) which is hiding this section for the installation page:

lovromazgon:
> There's still something I'd like to do (if possible) which is hiding this section for the installation page

@raulb try adding this to the top of the file:

```
--- 
# other options ...
hide_table_of_contents: true
---
```

lovromazgon (approved):
👏

raulb:
> > There's still something I'd like to do (if possible) which is hiding this section for the installation page
> 
> @raulb try adding [this](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#hide_table_of_contents) to the top of the file:
> 
> ```
> --- 
> # other options ...
> hide_table_of_contents: true
> ---
> ```

Nice thank you @lovromazgon. This actually plays out nicely in that page so we can have more real state for the terminal.

PR #89, opened 2024-05-09
update connectors' download links to link to our scarf package link

maha-hajja:
package created in scarf:
`conduit.gateway.scarf.sh/connector/download/{org}/{connector_name}/releases/download/v{version}/{connector_name}_{version}_{os_arch}.{ext}`

Fixes https://github.com/ConduitIO/conduit/issues/1549


src/connectorgen/main.go
```diff
@@ -16,13 +16,13 @@ package main
...
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"slices"
 	"strings"
 	"time"
 
-	"github.com/goccy/go-json"
```
hariso, line 25: Nitpick: it would be good to keep `goccy/go-json` (it was maybe an unintentional change), since we might have a large list of connectors, so we kinda care about performance.

hariso (approved):

PR #114, opened 2024-08-06
add schema utility functions to building.mdx

maha-hajja:
add docs for the Schemas access from processors, and the two Schema utility functions.

### DIFF
--- a/docs/processors/standalone/building.mdx
+++ b/docs/processors/standalone/building.mdx
@@ -291,14 +291,14 @@ Check [Compiling the processor](#compiling-the-processor) for what to do next, a
 
 ## Schemas
 
-Processors have access to the Schemas in the used Schema Registry using two utility functions:
-1. `schema.Create` : You can use this utility function to create a new Schema and add it to the Schema Registry.
+Processors have access to the schemas in the used Schema Registry using two utility functions:
+1. `schema.Create` : You can use this utility function to create a new schema and add it to the Schema Registry.
 This function can be called in any of the main processor methods.
     
 **Example**:
 ```go
 func (p *setProcessor) Open(ctx context.Context) error {
-	// Add a new Schema to the Schema registry before starting to process the records.
+	// Add a new schema to the schema registry before starting to process the records.
 	schemaBytes := []byte(`{
           "name": "record",
           "type": "record",
@@ -322,7 +322,7 @@ func (p *setProcessor) Open(ctx context.Context) error {
 }
 ```
 
-2. `schema.Get`: You can use this utility function to get a Schema from the Schema Registry using
+2. `schema.Get`: You can use this utility function to get a schema from the Schema Registry using
   its `version` and `subject`. This function can be called in any of the main processor methods.
 
   **Example**:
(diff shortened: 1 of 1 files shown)


docs/processors/standalone/building.mdx
```diff
@@ -289,6 +289,80 @@ func main() {
 
 Check [Compiling the processor](#compiling-the-processor) for what to do next, and how to compile the processor.
 
+## Schemas
+
+Processors have access to the Schemas in the used Schema Registry using two utility functions:
```
hariso, line 294: Q: is there a special reason for the capital S in `Schemas`?

docs/processors/standalone/building.mdx
```diff
@@ -289,6 +289,80 @@ func main() {
...
+```go
+func (p *setProcessor) Process(ctx context.Context, records []opencdc.Record) []sdk.ProcessedRecord {
+out := make([]sdk.ProcessedRecord, 0, len(records))
+for _, record := range records {
+    // get the schema subject name from the metadata
+    subject := rec.Metadata[opencdc.MetadataPayloadSchemaSubject]
+    // get the schema version from the metadata
+    version, err := strconv.Atoi(rec.Metadata[opencdc.MetadataPayloadSchemaVersion])
+    // get the schema 
+    get, err := schema.Get(ctx, subject, version)
```
hariso, line 338: There are helper methods such as `rec.Metadata.GetPayloadSchemaVersion()` (automatically converts the string to an int), and `rec.Metadata.GetPayloadSchemaSubject()`, so you don't have to work with the map and constants.

docs/processors/standalone/building.mdx
maha-hajja: not really no :/ I can change them to "schema", but should I keep "Schema Registry" capitalized?

hariso: I _think_ `Schema Registry` can stay capitalized (a name).

docs/processors/standalone/building.mdx
maha-hajja: did not know about these! thank you

docs/processors/standalone/building.mdx
```diff
@@ -289,6 +289,86 @@ func main() {
...
+## Schemas
+
+Processors have access to the schemas in the used Schema Registry using two utility functions:
+1. `schema.Create` : You can use this utility function to create a new schema and add it to the Schema Registry.
+This function can be called in any of the main processor methods.
+    
+**Example**:
+```go
+func (p *setProcessor) Open(ctx context.Context) error {
+	// Add a new schema to the schema registry before starting to process the records.
```
hariso, line 301: Nitpick: we have a helper for building schemas: https://github.com/ConduitIO/conduit-commons/blob/main/schema/avro/avro_builder.go#L37.

hariso (approved):
Great work!

docs/processors/standalone/building.mdx
maha-hajja: @hariso it took me some time to get that i needed to import two different `avro` packages to create the schema byte array (you need `"github.com/conduitio/conduit-commons/schema/avro"` and `avro2 "github.com/hamba/avro/v2"` to create something that looks like this:

> 	avro.NewBuilder("record", "records").AddField("int_field", avro2.NewPrimitiveSchema(avro2.Int, nil), avro2.WithDefault(100)).MarshalJSON()

I think this is a bit more complicated for the user to understand here and not the right place to document this method, WDYT? do we have it documented somewhere else in the docs?
I believe for the sake of the example here, the pure string is easier to understand, but we could link to the builder too.

hariso: You're right, and good that you're calling it out, we should rename the package then :+1:

PR #119, opened 2024-08-07
doc: configuration parameters

raulb:
Start collecting configuration parameters in a dedicated section.


docs/connectors/configuration-parameters/record-format.mdx
```diff
@@ -1,6 +1,6 @@
 ---
-title: "Output Formats"
-sidebar_position: 5
+title: "Record format"
```
lovromazgon, line 2: IMO we should still have "output" somewhere in the title, it makes it easier to find if you're trying to figure out how to write out the record in a specific format.

lovromazgon (commented):
It's a good start, I think we should add some examples, but that can also happen later on in a follow-up PR.

docs/connectors/configuration-parameters/record-format.mdx
raulb: @lovromazgon works for me

lovromazgon (approved):

PR #121, opened 2024-08-08
Fix broken links

hariso:


docusaurus.config.ts
```diff
@@ -18,7 +18,6 @@ const config: Config = {
   organizationName: 'conduitio', // Usually your GitHub org/user name.
   projectName: 'conduit-site', // Usually your repo name.
 
-  onBrokenLinks: 'warn',
```
lovromazgon, line 21: Don't we want to know when we have broken links?

hariso: The default is actually to throw an error and fail the build: https://docusaurus.io/docs/api/docusaurus-config#onBrokenLinks

lovromazgon: How about we make this explicit then and set it to `throw`?

lovromazgon: I mean that next time we look at this config we know that we wanted to have this option set explicitly and it's not there just because it's the default ;) Like documentation for our future selves, it was a conscious decision.

lovromazgon (approved):

Issue #120, opened 2024-08-07
doc: `skd.rate.*` as configuration parameter


raulb:
We started documenting some configuration parameters on [this page](https://conduit.io/docs/connectors/configuration-parameters/), but we're still missing `skd.rate.*` from this [list](https://github.com/search?q=repo%3AConduitIO%2Fconduit-connector-sdk%20%22sdk.&type=code) as of today.

PR #123, opened 2024-08-08
doc: add rate limit config param

raulb:
fixes https://github.com/ConduitIO/conduit-site/issues/120

### DIFF
--- a/docs/connectors/configuration-parameters/rate-limiting.mdx
+++ b/docs/connectors/configuration-parameters/rate-limiting.mdx
@@ -1,0 +1,47 @@
+---
+title: "Rate Limiting"
+sidebar_position: 2
+---
+
+Destination connectors can be configured to limit the rate at which records can be written. This is especially useful when the destination resource has a rate limit to ensure that the connector does not exceed it. By default, Conduit does not limit the rate at which records are written.
+
+## Configuration parameters
+
+* `sdk.rate.perSecond`: Maximum times records can be written per second (0 means no rate limit).
+* `sdk.rate.burst`: Allow bursts of at most X writes (1 or less means that bursts are not allowed). Only takes effect if a rate limit per second is set.
+
+## Example
+
+The pipeline will generate an infinite number of structured records in bursts during 5 seconds at a maximum rate, and write them to the log. When you run it, you'll notice that by using both `sdk.rate.perSecond` and `sdk.rate.burst`, the log destination connector will limit the rate of processed records to 1 burst per second.
+
+```yaml
+version: 2.2
+pipelines:
+  - id: generator-to-log
+    status: running
+    description: >
+      Example pipeline using the generator source connector and the log destination connector.
+      Showing how to limit the rate of processing records.
+    connectors:
+      - id: example
+        type: source
+        plugin: generator
+        settings:
+          rate: 0
+          recordCount: 0
+          burst.generateTime: 5s
+          burst.sleepTime: 0s
+          format.type: structured
+          format.options.id: int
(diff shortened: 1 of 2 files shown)


docs/connectors/configuration-parameters/rate-limiting.mdx
```diff
@@ -0,0 +1,47 @@
...
+title: "Rate Limiting"
+sidebar_position: 2
+---
+
+Destination connectors can be configured to limit the rate at which records can be written. This is especially useful when the destination resource has a rate limit to ensure that the connector does not exceed it. By default, Conduit does not limit the rate at which records are written.
+
+## Configuration parameters
+
+* `sdk.rate.perSecond`: Maximum times records can be written per second (0 means no rate limit).
+* `sdk.rate.burst`: Allow bursts of at most X writes (1 or less means that bursts are not allowed). Only takes effect if a rate limit per second is set.
```
lovromazgon, line 11: I was reading this and it made little sense to me, I checked the middleware and figured that it would make more sense to apply it _per record_ and not _per write_. The batch size shouldn't matter when configuring the rate limit IMO.

I opened this PR here, let's see what everyone thinks: https://github.com/ConduitIO/conduit-connector-sdk/pull/172

raulb: @lovromazgon makes sense to me. I'll change it here

docs/connectors/configuration-parameters/rate-limiting.mdx
```diff
@@ -0,0 +1,47 @@
...
+      Showing how to limit the rate of processing records.
+    connectors:
+      - id: example
+        type: source
+        plugin: generator
+        settings:
+          rate: 0
+          recordCount: 0
+          burst.generateTime: 5s
+          burst.sleepTime: 0s
```
lovromazgon, line 33: Since these parameters are optional and they are not the point of this example, I would simply remove them and let the generator produce records as fast as possible, the destination will then rate limit the whole pipeline.

docs/connectors/configuration-parameters/rate-limiting.mdx
```diff
@@ -0,0 +1,47 @@
...
+          operations: create
+      - id: log
+        type: destination
+        plugin: log
+        settings:
+          level: info
+
+          # limit the rate of proccessing records
+          sdk.rate.perSecond: 1
+          sdk.rate.burst: 1
```
lovromazgon, line 46: The burst also has no effect here, since it's the same as the rate per second. How about we change the example to have a rate per second of let's say 8 records and a burst of 2 records, then records should be arriving in the destination every 250ms in batches of 2 records at a time.

docs/connectors/configuration-parameters/rate-limiting.mdx
raulb: @lovromazgon the idea behind adding those even though they are optional was to illustrate than even when provided, they rate takes effect.

lovromazgon: My worry is that it will just confuse people 😕 besides, currently the burst is not doing anything, since sleep time is `0s` it means that the generator will simply generate records all the time, so removing these params should no change the outcome.

lovromazgon: Particularly the word `burst` in this parameters makes me think we should not include it in the example, because these might be confused with `sdk.rate.burst`, which have no connection with each other.

lovromazgon (approved):

PR #125, opened 2024-08-14
Clarify dead-letter queue behavior

lovromazgon:
I corrected the numbers in the examples.

### DIFF
--- a/docs/features/dead-letter-queue.mdx
+++ b/docs/features/dead-letter-queue.mdx
@@ -60,7 +60,8 @@ The first step is to change the window size and nack threshold:
 - If both the window size and the nack threshold are greater than 0, then nacked
   records will be sent to the dead-letter queue. Conduit will keep track of the
   number of nacks in the current window and stop the pipeline if the threshold
-  gets exceeded.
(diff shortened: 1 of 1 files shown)


docs/features/dead-letter-queue.mdx
```diff
@@ -49,18 +49,19 @@ By default, a Conduit pipeline will stop immediately when a record gets nacked.
 You can make the pipeline more fault-tolerant by enabling a dead-letter queue.
 The first step is to change the window size and nack threshold:
 
-- If the window size is greater than 1 and the nack threshold is 0, then the
+- If the window size is greater than 0 and the nack threshold is 0, then the
```
hariso, line 52: i.e. the window size doesn't matter in this case?

hariso (approved):

docs/features/dead-letter-queue.mdx
lovromazgon: It does, it has to be greater than 0, otherwise it's the second point, where records are sent to the DLQ, but the window is not monitored.

hariso: Ah, gotcha. :+1:

PR #128, opened 2024-08-28
feat: add additional links (and update dependencies)

raulb:
- Looking at how to disable a pipeline locally without having to comment the whole pipeline (I was interested in finding the `stopped` option for `status`, I realized the information wasn't where I was looking for so I add a link to cross reference one with another.
- Adds a link to the schema extraction configuration parameters to the schema support page.
- Update some dependencies.


maha-hajja (approved):
LGTM! nice work

PR #130, opened 2024-09-05
update to v0.11.1 instead of 0.11.0

maha-hajja:
noticed that the installation instructions still have v0.11.0 which wasn't stable, so updated to 0.11.1


hariso (approved):
Nice!

PR #138, opened 2024-09-19
Update guides for developing connectors

hariso:
Updates the guides for writing developing connectors with fresh examples from the file connector.

### DIFF
--- a/docs/connectors/building-connectors/developing-source-connectors.mdx
+++ b/docs/connectors/building-connectors/developing-source-connectors.mdx
@@ -13,10 +13,10 @@ Every Source implementation needs to include an [UnimplementedSource](https://pk
 
 ```go
 type Source struct {
-	sdk.UnimplementedSource
+    sdk.UnimplementedSource
 
-	config SourceConfig
-	tail   *tail.Tail
+    config SourceConfig
+    tail   *tail.Tail
 }
 ```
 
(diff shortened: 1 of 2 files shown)


docs/connectors/building-connectors/developing-source-connectors.mdx
```diff
@@ -74,128 +72,117 @@ func (s *Source) Parameters() map[string]sdk.Parameter {
...
-    ```go
-    func (s *Source) Open(ctx context.Context, pos sdk.Position) error {
-        configDirPath := s.config.Directory
-        files, err := ioutil.ReadDir(configDirPath)
-        if err != nil {
-            return fmt.Errorf("error reading directory '%s': %w", configDirPath, err)
-        }
+Every record read by a source connector has
+a [position](https://conduit.io/docs/features/opencdc-record#fields) attached.
+The position given to `Open()` is the position of the record that was last read
```
lovromazgon, line 92: Instead of "last read" I would say it's the last one that was successfully processed end-to-end, to make it clear that it's the _next_ record after this one that should be produced by the source.

docs/connectors/building-connectors/developing-source-connectors.mdx
```diff
@@ -74,128 +72,117 @@ func (s *Source) Parameters() map[string]sdk.Parameter {
...
+        return opencdc.Record{}, ctx.Err()
     }
-    ```
+}
+```
 
 ### `Ack()`
 
-Ack signals to the implementation that the record with the supplied position was successfully processed.
+`Ack` signals to the third party system that the record with the supplied position was successfully processed. It's worth that while some source connector need to implement this functionality (e.g. in the case of messaging brokers), others don't have to (e.g. a file source).
```
lovromazgon, line 166:
```suggestion
`Ack` signals to the third party system that the record with the supplied position was successfully processed. It's worth noting that while some source connectors need to implement this functionality (e.g. in the case of messaging brokers), others don't have to (e.g. a file source).
```

lovromazgon (approved):
A few suggestions, looks good otherwise 👍

docs/connectors/building-connectors/developing-source-connectors.mdx
hariso: Thanks, I'll change that now.

PR #141, opened 2024-09-29
feat: add configuration pipeline generator

raulb:
This is a POC that adds a pipeline generator to our documentation.

The idea is a user would select what source and destination would like to start with, an automatically generated configuration pipeline file will be shown. At that point, the user will be able to copy it and use it on their local Conduit. 

### In order to ship this to production, we'll need to:

1. Make sure we're able to automatically generate conduit YAML files with their specifications.
2. Update the way we generate the connector's JSON file. Notice we have three new keys for that purpose:

- To link to the path of the connector YAML  (`specificationPath`)
- To determine whether the connector has source configuration (`source`: true / false)
- To determine whether the connector has destination configuration (`destination`: true / false)

Recording of a demo can be found [here](https://meroxa.slack.com/archives/C07LNMU1ZR9/p1727623780216229).


raulb:
Closing for now. We'll revisit if priorities change.

PR #164, opened 2024-10-16
Link banner to changelog

lovromazgon:
Now that we have a changelog we can link to it rather than to github.


raulb (approved):
Great idea!

PR #174, opened 2024-10-21
fix: add redirect

raulb:
Original link referenced in https://github.com/ConduitIO/conduit/discussions/1724

### DIFF
--- a/docusaurus.config.ts
+++ b/docusaurus.config.ts
@@ -236,6 +236,10 @@ const config: Config = {
             to: '/docs/using/connectors/referencing',
           },
           {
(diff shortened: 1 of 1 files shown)


raulb:
While we're at it, I wonder whether this should be `using` or `developing`. Technically, this is a small change and not developing anything (aside from slightly modifying Conduit's source code to add a standalone connector as builtin), but curious to hear your thoughts @hariso @maha-hajja @lovromazgon

lovromazgon:
Both would fit, I don't have a strong opinion, but `using` seems fine.

lovromazgon (approved):

Issue #181, opened 2024-10-30
doc: Write a quickstart guide


hariso:

raulb:
I think this should be done now.

PR #183, opened 2024-10-31
Update the getting started guide

hariso:
Closes https://github.com/ConduitIO/conduit-site/issues/181.

### DIFF
--- a/docs/0-what-is/0-introduction.mdx
+++ b/docs/0-what-is/0-introduction.mdx
@@ -12,14 +12,14 @@ slug: '/'
   src="/img/conduit/on-white-conduit-logo.png"
 />
 
-Conduit is a data integration tool for software engineers, powered by [Meroxa](https://meroxa.io). Its purpose is to
-help you move data from A to B. You can use Conduit to send data from Kafka to
-Postgres, between files and APIs,
-between [supported connectors](/docs/using/connectors/list),
+Conduit is a data streaming tool for software and data engineers, powered
+by [Meroxa](https://meroxa.io). Its purpose is to help you move data from A to
+B. You can use Conduit to send data from Kafka to Postgres, between files and
+APIs, between [supported connectors](/docs/using/connectors/list),
 and [any datastore you can build a plugin for](/docs/developing/connectors/).
 
-It's written in [Go](https://go.dev/), compiles to a binary, and is designed to
-be easy to use and [deploy](/docs/installing-and-running?option=binary).
+It's written in [Go](https://go.dev/), compiles to a single binary, and is
+designed to be easy to use and [deploy](/docs/installing-and-running).
 
 Out of the box, Conduit comes with:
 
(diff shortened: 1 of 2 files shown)


docs/0-what-is/1-getting-started.mdx
```diff
@@ -5,81 +5,183 @@ sidebar_position: 0
...
-$ echo "hello conduit" >> example.in
-$ cat example.out
-hello conduit
+2024-10-30T13:31:08+00:00 INF component=plugin connector_id=example-pipeline:example-destination plugin_name=log plugin_type=destination record={"key":"c3RhbGxib2FyZA==","metadata":{"conduit.source.connector.id":"example-pipeline:example-source","opencdc.createdAt":"1730291468447408842","opencdc.payload.schema.subject":"example-pipeline:example-source:payload","opencdc.payload.schema.version":"1"},"operation":"create","payload":{"after":{"id":3312758023475813349,"name":"representativity"},"before":null},"position":"MQ=="}
 ```
 
-![scarf pixel conduit-site-docs-what-is](https://static.scarf.sh/a.png?x-pxid=01346572-0d57-4df3-8399-1425db913a0a)
\ No newline at end of file
+Here's just the generated record:
+
```
hariso, line 139: I'm thinking that it might be easier to see the data with a file destination. We would provide a `tail | jq` command, and a user would be able to see just the record. With the logs, it's a bit hidden.

docs/0-what-is/1-getting-started.mdx
```diff
@@ -5,81 +5,183 @@ sidebar_position: 0
...
+Now that you've got the basics of running Conduit and creating a pipeline
+covered, here are a few places to dive in deeper:
+- [Connectors](/docs/using/connectors/getting-started)
+- [Pipelines](/docs/using/pipelines/configuration-file)
+- [Processors](/docs/using/processors/getting-started)
+
+Looking for more examples? Check out the examples in
+our [GitHub repository](https://github.com/ConduitIO/conduit/tree/main/examples).
+
+<!-- TODO add scarf pixel -->
```
hariso, line 187: I'll do this in a bit.

docs/0-what-is/1-getting-started.mdx
```diff
@@ -5,81 +5,183 @@ sidebar_position: 0
...
+in [our installation page](/docs/installing-and-running).
+
+:::note
+The Conduit binary contains both, the Conduit service and the Conduit CLI, with
+which you can interact with Conduit.
 :::
 
-Create a folder called `pipelines` at the same level as your Conduit binary.
-Inside of that folder create a file named `file-to-file.yml`.
+## Initialize Conduit
```
hariso, line 36: Since this is a quick start, we might even want to skip this part. It's optional anyway, and we can cut the time to getting the pipeline running.

docs/0-what-is/1-getting-started.mdx
```diff
@@ -5,81 +5,185 @@ sidebar_position: 0
...
+## Run Conduit
+
+With the pipeline configuration being ready, we can run Conduit:
 
 ```shell
-$ ./conduit
-/.../
-2023-04-01T12:34:56+00:00 INF pipeline configs provisioned component=provisioning.Service created=["file-to-file"] deleted=[] pipelines_path=./pipelines
-/.../
+$ conduit --pipelines.path pipelines/
```
hariso, line 127: Needed, because Conduit is installed in another directory.

docs/0-what-is/0-introduction.mdx
```diff
@@ -12,264 +12,40 @@ slug: '/'
...
+Conduit is a data streaming tool for software and data engineers, powered
+by [Meroxa](https://meroxa.io). Its purpose is to help you move data from A to
+B. You can use Conduit to send data from Kafka to Postgres, between files and
+APIs, between [supported connectors](/docs/using/connectors/list),
 and [any datastore you can build a plugin for](/docs/developing/connectors/).
 
-It's written in [Go](https://go.dev/), compiles to a binary, and is designed to
-be easy to use and [deploy](/docs/installing-and-running?option=binary).
+It's written in [Go](https://go.dev/), compiles to a single binary, and is
+designed to be easy to use and [deploy](/docs/installing-and-running).
```
raulb, line 22: now that I see this link, I don think we should link `deploy` to this page. It's about installing, so I don't think it makes sense at all.

docs/0-what-is/0-introduction.mdx
```diff
@@ -12,264 +12,40 @@ slug: '/'
...
+* **Easily extensible**: Conduit connectors are plugins that communicate with Conduit
+  via a gRPC interface. This means that plugins can be written in any language
+  as long as they conform to the required interface. Check out
+  our [connector docs](/docs/using/connectors/getting-started)!
 
-Eliminate the multi-step process you go through today. Just download the binary
-and start building. Conduit pipelines run in their own goroutines and are
-connected using Go channels. This makes Conduit pipelines incredibly performant
-on multi-core machines.
+The full list of features can be found on
```
raulb, line 46: I don't think this is accurate, since these are _other_ features that aren't highlighted as first elements on the main navigation. I would probably rephrase this to something like

`For a detailed look at additional features, visit our [Features](/docs/using/other-features) page. Here, you'll find a selection of tools and functionalities that complement our main features.`

docs/0-what-is/1-getting-started.mdx
```diff
@@ -5,81 +5,191 @@ sidebar_position: 0
...
+:::note
+The Conduit binary contains both, the Conduit service and the Conduit CLI, with
+which you can interact with Conduit.
 :::
 
-Create a folder called `pipelines` at the same level as your Conduit binary.
-Inside of that folder create a file named `file-to-file.yml`.
+## Initialize Conduit 
+
+First, let's create a directory where we will build our first pipeline:
```
raulb, line 34: I'd personally avoid this step. Let's assume people know where to create it. In the worst-case scenario, they delete the stuff where it's created, and then they'll do it again in the right place.

docs/0-what-is/1-getting-started.mdx
raulb: I would leave it with just the option of `conduit init` to be honest. I think it's a good idea to guide them through that since we have that command available. It'll help with educating them about where things are.

docs/0-what-is/1-getting-started.mdx
```diff
@@ -5,81 +5,191 @@ sidebar_position: 0
...
+$ conduit pipelines init
 ```
 
-Copy following content into `file-to-file.yml`:
+`conduit pipelines init` builds an example that generates flight information
+from an imaginary airport every second. Use `conduit pipelines init --help` to
+learn how to customize the pipeline.
+
+If the `pipelines` directory, you'll notice a new file,
+`pipeline-generator-to-file.yaml` that contains our pipeline's configuration:
```
raulb, line 76: I would add a call out as 

```
:::tip 

Here's a full detailed information about Pipeline configuration files (/docs/using/pipelines/configuration-file)

:::
```

docs/0-what-is/1-getting-started.mdx
raulb: Since we initialized this directory via `conduit init`, shouldn't this path be known? (i.e.: no need to specifying the flag)

docs/0-what-is/1-getting-started.mdx
```diff
@@ -5,81 +5,191 @@ sidebar_position: 0
...
+{
+    "airline": "wheelmaker",
+    "scheduledDeparture": "2024-11-05T10:06:34.148469Z"
+}
 ```
 
-![scarf pixel conduit-site-docs-what-is](https://static.scarf.sh/a.png?x-pxid=01346572-0d57-4df3-8399-1425db913a0a)
\ No newline at end of file
+The pipeline will keep streaming the data from the generator source connector to
+the file destination connector as long as Conduit is running.
```
raulb, line 182: Mention how to stop it, maybe?

raulb (commented):
@hariso nice work. Only some suggestions. Let me know if they make sense to you.

docs/0-what-is/0-introduction.mdx
hariso: Makes sense!

docs/0-what-is/0-introduction.mdx
hariso: Agreed, I'll change it to `install`.

docs/0-what-is/1-getting-started.mdx
hariso: I think you're right... I thought I'll need that path somewhere, but I didn't.

docs/0-what-is/1-getting-started.mdx
hariso: We have that under "What's next?" or do you think we should move it here?

docs/0-what-is/1-getting-started.mdx
hariso: I thought so too, but nope, it's not. If you install Conduit via a package manager, then you'll use that one binary whenever you want to run Conduit, regardless of what's your working directory. That means that you always need to give it the path. For example, on my machine, Conduit is at `/usr/bin/conduit`. So we either need the `pipelines.path` flag or the config path.

Another option is to change Conduit so that it always scans the current working directory. But I'm not sure that's the best way when Conduit is working as a service in a VM or in a container, when I think it should always use a known path.

docs/0-what-is/1-getting-started.mdx
hariso: Yup! :+1:

docs/0-what-is/1-getting-started.mdx
```diff
@@ -5,81 +5,190 @@ sidebar_position: 0
...
+:::note
+The Conduit binary contains both, the Conduit service and the Conduit CLI, with
+which you can interact with Conduit.
 :::
 
-Create a folder called `pipelines` at the same level as your Conduit binary.
-Inside of that folder create a file named `file-to-file.yml`.
+## Initialize Conduit 
+
+Let's initialize the "working environment":
```
raulb, line 34:
```suggestion
Firs, let's initialize the working environment:
```

docs/0-what-is/1-getting-started.mdx
raulb: @hariso you're right. That's enough.

docs/0-what-is/1-getting-started.mdx
```diff
@@ -5,81 +5,190 @@ sidebar_position: 0
...
+    "airline": "wheelmaker",
+    "scheduledDeparture": "2024-11-05T10:06:34.148469Z"
+}
 ```
 
-![scarf pixel conduit-site-docs-what-is](https://static.scarf.sh/a.png?x-pxid=01346572-0d57-4df3-8399-1425db913a0a)
\ No newline at end of file
+The pipeline will keep streaming the data from the generator source connector to
+the file destination connector as long as Conduit is running. To stop Conduit,
+press `Ctrl + C` (on a Linux OS, or the equivalent on other OSes). This will
```
raulb, line 178:
```suggestion
press `Ctrl + C` (on a Linux OS, or the equivalent on other operating systems). This will
```

raulb (commented):
@hariso just a couple of nits. Looks good.

Issue #167, opened 2024-10-17
doc: Generate documentation page for each connector


lovromazgon:
Once we implement the [connector.yml](https://github.com/ConduitIO/conduit/issues/1523) feature we can crawl through all connectors in the connector list, collect the YAML files and generate the documentation page for each connector. We should be able to create a filtered list of connectors that we want to generate pages for, to prevent us from creating the documentation for any third-party created connector.

PR #200, opened 2024-12-06
Generate connector docs

hariso:
Closes https://github.com/ConduitIO/conduit-site/issues/167.

Here's how the updated connectors' list and a new connector's page look like:

**Connectors list** (and search)

Screenshot from 2024-12-20 20-48-37

**A generated page**

Screenshot from 2024-12-20 20-47-47


static/connectors.json
```diff
@@ -12378,15 +12563,462 @@
...
     "description": "Conduit connector for Weaviate",
+    "conduitIODocsPage": "",
     "createdAt": "2023-07-13 07:27:26 +0000 UTC",
     "url": "https://github.com/devarispbrown/conduit-connector-weaviate",
     "stargazerCount": 0,
     "forkCount": 0,
     "releases": []
   },
+  {
+    "nameWithOwner": "hariso/conduit-connector-foo",
```
hariso, line 12574: Need to remove this.

static/connectors.json
```diff
@@ -12378,15 +12563,462 @@
...
+            "download_count": 0,
+            "size": 6571106
+          }
+        ],
+        "is_latest": false
+      }
+    ]
+  },
+  {
+    "nameWithOwner": "hariso/conduit-connector-kafka",
```
hariso, line 12885: Need to remove this.

raulb (commented):
@hariso on https://github.com/ConduitIO/conduit-site/pull/203/files#diff-7a45f875fcea055cb151d9ae10b11385301584ae2bd1a9b111ea0b6c8e307c66, I went through the process of removing the artifacts and only link to the latest release which means we would be able to remove the `IsLatest` release check. I believe we should simplify that bit, and only show the relevant information there.

Another suggestion I have is to use the name of the connector as a link instead of expanding the accordion of releases, eliminating the `View docs` link.

What do you think?

src/connectorgen/generate_docs.go
```diff
@@ -0,0 +1,29 @@
+// Copyright © 2024 Meroxa, Inc.
```
raulb, line 1: 😂

```suggestion
// Copyright © 2025 Meroxa, Inc.
```

raulb (approved):
Looks great! Looking forward to seeing the dedicated pages once https://github.com/ConduitIO/conduit/issues/2054 ships.

src/connectorgen/generate_docs.go
hariso: No, don't do that to me!

PR #203, opened 2024-12-13
feat: add a doctor page to indicate statuses for all our connectors

raulb:
This is my champagne project for the Q4 online offsite we just held at Meroxa.

Internally, the Conduit team has kept a list of checks we've been manually reviewing to ensure all our connectors are up to speed when it comes to using the latest versions of our libraries. We also want to make sure those contain certain workflows and tasks so that all of them maintain some cohesion and standards.

With this pull request, my intention is to have an automated and easier way to identify where our attention needs to be dedicated.

It relies on the existing workflow we run every week to update our connector list inventory.

⚠️ To reduce the number of pre-existing GitHub requests, I decided to avoid fetching anything that's not a latest release since that's the information I would consider relevant. For anything else, I think it should be checked at their GitHub repository directly.

Some improvements I'd like to see in the future:

1. Have a mild status in which certain checks are neither green nor red, but maybe orange. This would be useful, for example, to indicate if a connector is maybe not using the latest patched version, but it is using the same major and minor versions.
2. Update URLs based on filters so they could be shared.
3. Maybe include some other fields such as the number of issues or pull requests opened.

To see this, just visit this URL: https://conduit-doctor.conduit-site.pages.dev/doctor

Demo: https://www.loom.com/share/0dad78b4175f429087aea3d7e9e6ba43


raulb:
This was part of a champagne project. Closing for now 😢 .

PR #215, opened 2025-01-13
Exclude SDK from dependents

hariso:
Test modules within the SDK use SDK as a dependency, so the SDK is included as a dependent of itself.


lovromazgon (approved):
👍

PR #222, opened 2025-01-28
Connector specification registry

lovromazgon:
This PR refactors connectorgen so that it now has 3 subcommands:

- `registry` generates `connectors.json`, it fetches all releases and assets, the resulting file serves as a registry of all available connectors. The change here is that we again have all releases in the file, not only the latest one.
- `specifications` iterates over the connectors in the JSON file and tries to download `connector.yaml` from the root of the repo. The specifications file is stored in `static/connectors/github.com/{owner}/{repo}/connector.yaml`. The tool also creates a file called `.metadata.yaml` in the same folder, which stores the commit SHA at which the repo was checked for the `connector.yaml` and the time it was checked - this serves as the information if the file could have potentially changed and we need to download it again.
- `pages` generates the documentation page for any connector in the `connectors.json` file that has at least one release with a `connector.yaml` file. It is using [readmegen](https://github.com/ConduitIO/conduit-connector-sdk/pull/115) behind the scenes to generate the pages.

Note that most of the code was simply refactored and reused. It doesn't look like it though, because it was moved to new files to have one source file per command.

As part of this PR the CI action, makefile, documentation page template and connector list page were updated to match the changes, however the resulting pages still look the same.

Oh, and the last commit on the PR contains the result of running `make generate` in connectorgen, something that will otherwise be done on a weekly basis in the CI action.


src/components/ConnectorList/ConnectorAccordion/index.tsx
```diff
@@ -15,36 +15,40 @@ import ReleaseAccordion, { Release } from '@site/src/components/ConnectorList/Re
 import { Paper } from '@mui/material';
 
 export class Connector {
-  nameWithOwner: string;
+  name_with_owner: string;
```
raulb, line 18: Why changing these to  snake case?. Camel case is more common in JS / React. Personally, I'd like to see these as camel case.

raulb (commented):
Looks good. Just one comment about styling.

src/components/ConnectorList/ConnectorAccordion/index.tsx
lovromazgon: I personally don't care so much one way or the other, I changed it to make it consistent, because we mixed them both (see nested structs). If you feel strongly I can go in and change the nested structs instead 🤷

lovromazgon: One thing I'll add is that GitHub uses snake case in their API ([example](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#get-a-repository)), and since this data comes directly from their API, keeping it in snake case would better align with their data model. I'm trying to avoid extra work here 🙈

raulb (approved):

PR #235, opened 2025-02-04
Connector docs template fix

lovromazgon:
Fixes the template so it doesn't fail if a connector does not implement a source or destination.

Also connectorgen was ran in this PR so it generated a few additional pages and fetched more connector.yaml files.


raulb (approved):
LGTM

PR #258, opened 2025-03-20
Benchi announcement changelog

lovromazgon:
This adds a changelog for Benchi, similar to the blog post, but more concise.


raulb (approved):
🚀

PR #270, opened 2025-04-22
Allowed/denied connectors in registry

lovromazgon:
Change which connectors get added to the connector registry.

Previously, we added any repository that imported the connector SDK, regardless of the org. Now the process has been improved to give us more control over what we add to the list and prevent malicious actors from adding malicious connectors to our list. Only connectors that are from allowed orgs are added automatically (`conduitio`, `conduitio-labs` and `meroxa`), all others have to be explicitly added to the list.

To make it possible for us to still spot new 3rd party connectors, I added code that exports the denied connectors into `src/connectorgen/denied-connectors.json`. Note that this file is not in the `static` folder, so as not to advertise them on our page (the `src` folder is not included on our page). This file is meant only for us to notice connectors that mistakenly land in the denied list.

Actual changes to the code:

  - The connector list was moved to a `yaml` file, so it's simpler to edit
  - The registry command has a new flag `-d`/`--denied-path` that, if supplied, exports the denied list to a file (otherwise it skips this step)
  - The code was refactored a bit (using `ghdeps.Repository` instead of a `string` for representing a repo)

This PR also contains the regenerated `connectors.json` file, superseding https://github.com/ConduitIO/conduit-site/pull/269.

> [!NOTE]
> I removed all repositories from the list that were forks or ones that were eventually moved to conduitio-labs. I only left 3rd-party connectors, which are actually different from our organization's repositories.


src/connectorgen/registry-config.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - lovromazgon/conduit-connector-kafka-broker
+  - nickchomey/conduit-connector-surrealdb
+
+# The deny list is used to exclude specific repositories from the allow list.
+# The deny list takes precedence over the allow list.
+# The deny list can either contain a string that matches the repository name
+# or an organization name with a wildcard (*) as the repository name.
+# The comparison will be case-insensitive.
+deny:
+  # Deny specific repositories that are allowed in the orgs above
```
raulb, line 22: Could we, instead of individually denying specific repositories, maybe only allow those that follow the format:

  - ConduitIO/conduit-connector-*
  - conduitio-labs/conduit-connector-*
  - meroxa/conduit-connector-*

This way, we wouldn't need to worry about manually updating this list on any other repository we might (we will) create in the future on any of these orgs.

raulb (commented):
Thanks @lovromazgon for taking care of this.

Left a comment below which is related to something else I thought when I looked at the denied-connectors.json. I think we should update the criteria to _what's considered a connector_ (which I think we should standardize as `conduit-connector-*`) to make our lives easier. 

It seems like we we have exceptions to make on repositories that don't look like a connector (e.g.: `raulb/custom-conduit`). 

What I was thinking was:

Allow list:

ConduitIO/conduit-connector-*
conduitio-labs/conduit-connector-*
meroxa/conduit-connector-*

Individual exceptions (added manually)

_Everything else is denied_

What do you think?

src/connectorgen/registry-config.yaml
lovromazgon: Good idea! I'll add the change 👍

lovromazgon: See https://github.com/ConduitIO/conduit-site/pull/270/commits/debb06d465c86599d070a758f794b6b7913a484d 👍

raulb (commented):
@lovromazgon looks great!

Only one comment before approving. It is my understanding that with the recent changes, technically speaking, we would only need to add to the list of `defined-connectors.json` https://github.com/ConduitIO/conduit-site/pull/270/commits/debb06d465c86599d070a758f794b6b7913a484d#diff-28f0f12ca05137174cd42d88e1d2e40ba4393941a5b470f8759546873d49e00fR12, right?

If we want to explicitly call out some contributors we could, but I don't think that's needed? We could also remove other owned repositories such as `ConduitIO/conduit`, `ConduitIO/conduit-operator`...

raulb (approved):

raulb:
> @lovromazgon looks great!
> 
> Only one comment before approving. It is my understanding that with the recent changes, technically speaking, we would only need to add to the list of `defined-connectors.json` `ConduitIO/conduit-connector-template`, right?
> 
> If we want to explicitly call out some contributors we could, but I don't think that's needed? We could also remove other owned repositories such as `ConduitIO/conduit`, `ConduitIO/conduit-operator`...

Update: @lovromazgon and I talked about this today after standup, and clarified my comments. Approved 🚀

Issue #282, opened 2025-05-09
Improve header on connector pages in dark mode


lovromazgon:
On a connector page (e.g. https://conduit.io/docs/using/connectors/list/file) we have a header on top that lists the author, the link to GitHub and the number of stars.

In light mode, that header has a grey background, the divider and the number of GitHub stars are nicely visible:

![Image](https://github.com/user-attachments/assets/0ef0ec37-0552-49e0-ac09-d212dbe5c284)

In dark mode this could be improved, as the background and divider are basically invisible, while the stars are barely visible.

![Image](https://github.com/user-attachments/assets/279eab8f-88df-41c3-9a82-d8b27f11a351)

PR #285, opened 2025-05-13
fix: Connector header on dark mode

raulb:
Fixes https://github.com/ConduitIO/conduit-site/issues/282

**Before**

CleanShot 2025-05-13 at 3  54 07@2x

**After**

CleanShot 2025-05-13 at 4  01 05@2x


lovromazgon (approved):
❤️

PR #286, opened 2025-05-13
doc: Record size limit on standalone connectors

raulb:
Related https://github.com/ConduitIO/conduit/issues/547.

### DIFF
--- a/changelog/2025-05-xx-max-limit.mdx
+++ b/changelog/2025-05-xx-max-limit.mdx
@@ -1,0 +1,16 @@
+---
+slug: '2025-05-xx-max-limit'
+title: New flag...
+draft: false
+tags: [conduit, cli]
+---
+
+Add reference to the [updated doc](/docs/using/connectors/installing#record-size-limitations).
+
+<!--truncate-->
+
+### Key Highlights
+
+
+
+![scarf pixel conduit-site-changelog](https://static.scarf.sh/a.png?x-pxid=b43cda70-9a98-4938-8857-471cc05e99c5)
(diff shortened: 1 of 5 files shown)


docs/1-using/7-other-features/5-api.mdx
```diff
@@ -9,8 +9,16 @@ Conduit exposes a gRPC API and a REST API.
...
 using the CLI flag `-grpc.address`. To learn more about the gRPC API please have
 a look at
-the [protobuf file](https://github.com/ConduitIO/conduit/blob/main/proto/api/v1/api.proto)
-.
+the [protobuf file](https://github.com/ConduitIO/conduit/blob/main/proto/api/v1/api.proto).
+
+:::info
+Conduit uses gRPC streams to communicate with [standalone connectors](/docs/core-concepts#standalone-connector).
+
+Make sure to keep the [api configuration](/docs/configuration) enabled (`--api.enabled`) if you want to use [standalone connectors](/docs/core-concepts#standalone-connector).
```
lovromazgon, line 17: Wait, why? The connectors shouldn't be using the user facing API, and the connector utilities should start regardless of the flag ([utils start here](https://github.com/ConduitIO/conduit/blob/0460da9e30237e57620c5c5ca7da2f55c5588e4d/pkg/conduit/runtime.go#L818), while flag is checked [here](https://github.com/ConduitIO/conduit/blob/0460da9e30237e57620c5c5ca7da2f55c5588e4d/pkg/conduit/runtime.go#L345C18-L345C25)). If that's not the case, that's definitely a bug.

raulb: @lovromazgon that's true! When reading the code for some reason I assumed that was the case. Didn't realize these are completely separate. Will fix.

hariso (approved):

PR #291, opened 2025-05-20
add rabbitmq v0.4.0 to the changelog

alarbada:
Adds rabbitmq v0.4.0 to the changelog.

Note for the reviewer, I don't know if there's a project command to automate link generation for new versions of a given connector? I did a quick find and replace from 0.3.0 to 0.4.0 for rabbitmq, but there's some folders (static/connectors/github.com) that make me suspicious of some existing automation?


raulb (approved):
Looks great, thank you!

PR #318, opened 2025-06-11
update docs for `pipelines init`

raulb:
Related PR https://github.com/ConduitIO/conduit/pull/2331 + https://github.com/ConduitIO/conduit/pull/2333

> [!NOTE]  
> This should go live as part of the 0.14.0 release since that's when we're releasing the Conduit changes.


hariso (approved):

raulb:
Need to update it once https://github.com/ConduitIO/conduit/pull/2333 ships

PR #351, opened 2026-07-23
docs: scaffold with `new` + registry index-schema reference

devarismeroxa:
## Summary

- New how-to: **Scaffold a New Connector or Processor** (`docs/2-developing/0-connectors/2-scaffold-new.mdx`) — documents `conduit connector new` / `conduit processor new`: what each step generates, the shared flags, `--json` output shape, and the stable `scaffold.*` error codes. Explicitly reconciles the new `new` verb with the existing [Conduit Connector Template](https://github.com/ConduitIO/conduit-connector-template) page instead of duplicating it: `new` is framed as the fast CLI path that *drives* the same template, the template page as the manual/GitHub-first path. Cross-linked from both the connector and processor template pages and their section intros.
- New reference: **Registry Index Schema** (`docs/1-using/5-connectors/3-registry-index-schema.mdx`) — exhaustive field-by-field reference for the signed connector-registry index document: envelope (`payload` + `signatures[]`), verification order, every payload/connector/version/artifact field, the root-vs-freshness signature roles and trust-anchor model, and the `registry.*` error codes. Aimed at self-hosted operators pointing `--index-url`/`--index-file` at their own index and at anyone building tooling against the format.
- Renumbered `docs/2-developing/0-connectors/2..8` up by one to slot the new scaffolding page in right after the template page (pure renames, zero content diff — verified with `git show --stat`). Routes are slug-based, not number-based, so no existing links break.

## Grounding

Verified against source, not the aspirational design doc, where they diverge:

- `cmd/conduit/root/connectors/new.go`, `cmd/conduit/root/processors/new.go`, `cmd/conduit/internal/scaffoldcmd/newcmd.go`, `pkg/scaffold/{request,codes,preflight,doc}.go`, `pkg/scaffold/steps/*.go` — the actual `Flags` struct only has `--lang --module --path --sdk-version --git/--no-git --skip-generate --yes --force`. The design doc's `--template-ref/-repo`, `--remote`, and TTY-prompt flow are **not implemented** — the page says so plainly rather than documenting flags that don't exist yet.
- `pkg/registry/index/{schema_v1,envelope,codes,freeze,anchors}.go`, `docs/design-documents/20260714-connector-registry-index-schema.md`, `docs/design-documents/registry-index/index-schema.json`, and `pkg/registry/index/testdata/sample-index.json` for the worked example.
- `cmd/conduit/root/connectors/install.go` for the honest self-hosting caveat: trust anchors are compiled in at build time, so pointing `--index-url` at a self-signed private index fails closed with `registry.trust_anchor_expired` on a stock binary — there's no supported way to add a custom trust anchor to a stock build. Flagged explicitly as a `:::caution` rather than glossed over.

## Unverified / not run

- Did not run a full `yarn build` (no `node_modules` in this environment) — verified MDX validity manually instead: balanced code fences, no bare `<...>` outside code spans/fences, front matter matches the site's `title:`-only convention. Recommend CI's Docusaurus build check confirms this before merge.
- Did not check whether `docs.conduit.io`'s markdownlint config (if any) enforces a line-length rule that would flag the reference page's wide table rows — no `.markdownlint*` config was found in this repo, so this wasn't a blocker, but flagging in case CI has a stricter external rule.

## Test plan

- [ ] Docusaurus build passes in CI
- [ ] Spot-check both new pages render (tables, code fences, internal links resolve)
- [ ] Confirm the renumbered `0-connectors` section still sorts correctly in the sidebar

https://claude.ai/code/session_015GQFzakPShAYj8CcwajYDD


devarismeroxa:
Superseded by #352 (consolidated v0.17+v0.18 docs — resolves sidebar-numbering across all 8 drafts, build-verified).

